PR #85, opened 2024-07-09
Tear Down Class, Cleanup after test.py

hudsonwhipple:
Test.py
- Teardown class added
- Calls to deletion endpoints

User.py, two new endpoints
- delete_instructor
- delete_student

Documentation for endpoints

### DIFF
--- a/backend/test.py
+++ b/backend/test.py
@@ -1,20 +1,11 @@
 import unittest
 import requests
 
-from routes.assignment import delete_assignment
-from routes.course import delete_course
-from routes.user import delete_instructor, delete_student
-
 env_file = "http://localhost:5000"
-
-
 
 class APITestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        # print(dir(cls))
-        # app.config['TESTING'] = True
-        # cls.client = app.test_client()
 
         """Create test data that will be used across multiple tests."""
         cls.instructor_id = cls.create_instructor()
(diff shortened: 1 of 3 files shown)


backend/test.py
```diff
@@ -1,10 +1,12 @@
 import unittest
 import requests
 
+env_file = "http://localhost:5000"
```
defkorean, line 4: We could probably add this environmental variable to the .env file in backend/.env since other developers may have a different url

defkorean:
While trying to run the test script, I run into an error where there's invalid json data. I've tried running in virtual environments and local environment and deleting/adding back docker containers but it produces the same traceback here:
Screenshot 2024-07-09 at 12 15 59 PM

If anyone else is running into this or has a fix please let me know. Otherwise the code and documentation itself looks good, I just can't run the script.

Vellankijay (changes requested):
Hey Hudson, the implementation as is on the original code is working. However, with my changes to the database and centralizing all of the users under one table needs one simple revision. If you need assistance, please have a look at my branch jay_vellanki. 

Example of how the people's table changes logging in people:
Screenshot 2024-07-18 at 12 15 28 PM


The only changes you have to make to your code is in user.py and test.py. In user.py, just implement one delete method for all people. This is simple, can be done by replacing all instructor related table calls with people's table. As for the test.py, just write a test case testing the method.

defkorean:
> While trying to run the test script, I run into an error where there's invalid json data. I've tried running in virtual environments and local environment and deleting/adding back docker containers but it produces the same traceback here: Screenshot 2024-07-09 at 12 15 59 PM
> 
> If anyone else is running into this or has a fix please let me know. Otherwise the code and documentation itself looks good, I just can't run the script.

It turns out that this issue can be reproduced if you ran the test script before these changes. After deleting all the rows to get a fresh database, I was able to run the test script multiple times. Other than the proposed changes above, I think it's good to merge.

kiat:
We do this PR after changes to the database structure.

Issue #79, opened 2024-06-21
Manage Submissions and Assignment results (instructor end) display error


TanishG123:
**Problem:**
Due to poor stage management, the submission data (on manage submissions) and the student's name (on instructor's end) sometimes disappear on reload.

**Solution:**
Implemented better information fetching on render to always display the correct information. (Still testing, will be added to the resubmit feature pull request)

**Files Changed:**
manageSubmissions/index.js
results/index.js

Issue #87, opened 2024-07-09
Docker Setup on Assignment Creation Cleanup (Prevent New Docker Container Creation on Each Submission)


TanishG123:
**Issue:**
Creating a new Docker container on each submission leads to prolonged runtimes for submissions. We need to streamline the process to enhance performance and efficiency.

**Proposed Solution:**
- **Single Docker Container Creation:** 
  - A single Docker container should be created when the instructor creates the assignment.
  - This container should be updated when the instructor uploads the autograder zip file.
  - The container should be built and executed with the zip upload, maintaining a structured file system within the container to handle different students, their submissions, and results.

- **Assignment Table Update:**
  - Add a column in the Assignment table to store the container ID.
  - Upon container creation, store the container ID in the database.
  - Access this container ID from the database to manage and run the container as needed.

- **Container Execution on Submission:**
  - On submission, the pre-existing container must be utilized to run the autograder.
  - The container should handle the submission, executing the same processes as currently implemented.

- **Error Handling:**
  - Implement checks to handle/prevent errors such as:
    - Container not created.
    - Autograder not uploaded.
    - Container corruption.
    - Container deletion.
  - Ensure that an autograder file is provided before publishing an assignment.
  - Implement mechanisms to recreate containers and restore their data as necessary in case of corruption.

**Locations/Files:**
- TBD

TanishG123:
V1 implemented with PR #90. This is a basic implementation that needs consideration on how to handle container damage or deletion.

DarioDeSilva:
Status update: image reuse was implemented in PR #90 — the autograder image is built once at zip-upload time and stored on the assignment (autograder_image_name), rather than rebuilding per submission. However, a new container is still created and torn down on every submission, and there's minimal handling for a missing/corrupted image at submission time. Leaving open, re-scoped to container reuse + corruption handling.

PR #90, opened 2024-07-11
Resubmit Feature and Issue #79, 87

TanishG123:
**Description**
- Added a Resubmit Feature that allows students to submit their files for an assignment multiple times. 
- Autograder runs on each new submission and logs the results to the DB as well
- Implemented a Submission History Modal that gives the student the ability to view all of their submissions for this assignment
- Implemented the notion of active submission. The active submission is what the instructor sees for grading. Students can assign active submissions in the submission history modal.
- The student can only leave regrade requests on active submissions
- If a student tries to resubmit or activate an old submission while the current active submission has a regrade request they will be prompted if they want to delete that regrade request and continue or not.
- The rest of the grade request functionality is the same.
- Fixed errors in deleting courses and assignments to ensure dependencies in regrade requests and enrollments are removed.
- Added barebones functionality for container setup and persistence on assignment auto-grader upload. Execution depends on the fact that the container isn't deleted on the local system and no recovery methods are implemented in case of deletion (**future consideration**). The feature is functional and appears to execute significantly faster on assignment submission. However autograder upload takes a little more time because of the possibility of building a new docker image + container. The feature is open for feedback on implementation, functionality, or future changes. More details can be found in the Issue.


TanishG123:
Hey guys I also added in the "Docker Setup on Assignment Creation Cleanup (Prevent New Docker Container Creation on Each Submission)" Issue #87. More details are in the comment or the Issue itself and while it's functional it's not final so feel free to leave any comments.

backend/routes/assignment.py
defkorean, line 1: Looks good to me, good addition to make sure to remove those regrade requests alongside the submission.

backend/routes/course.py
```diff
@@ -97,6 +97,12 @@ def delete_course():
     if related_assignments:
         print("410")
         return jsonify("Assignments must be deleted"), 410
+    
+    enrollments = db.session.query(Enrollment).filter_by(course_id=course_id).all()
+    if (enrollments):
```
defkorean: For consistent style, I suggest removing the parentheses. Otherwise, it appears the same logic exists here as in assignments.py so 👍

backend/routes/submission.py
defkorean, line 1: I would remove commented out code if it's functional as before (i.e I suggest creating a unit test for these calls). Additionally, document these new calls in docs/CodeAssist_backend_endpoints.md

frontend/src/pages/assignments/assignment_modal.js
defkorean, line 1: If the assignment modal is opening as intended, I think you can either choose to remove the console logs or comment them out since we don't want these prints to show up when it goes into production

frontend/src/pages/result/StudentInfoPanel.js
```diff
@@ -17,47 +18,24 @@ export default function StudentInfoPanel({
   const [EditGradeModalVisible, setEditGradeModalVisible] = useState(false);
   const [Grade, setGrade] = useState("");
   const [Justification, setJustification] = useState(null); // Initialize as null
-  const { assignmentId, studentId } = useParams();
+  //const { assignmentId, studentId } = useParams();
```
defkorean: Remove this line entirely

defkorean (dismissed):
Overall, really great additions to the codebase. Please review my comments found in some of these files (just minor changes) and I believe it would be good to be merged.

defkorean (approved):
All good to merge!

kiat:
Thanks!

PR #93, opened 2024-07-18
Ticket 84 - Prevent Duplicate Assignment Names

braden-carter:
We want to prevent instructors from creating assignments with 2 names that are the same. 

**backend\routes\assignment.py**

We altered the create_assignment() method so that it returns an error 400 with a message, should a duplicate assignment name be found in the same course.

**frontend\src\pages\instructor\assignments\CreateAssignment.js**

Has a variable that now turns the assignment name box red should an error be set

**frontend\src\pages\instructor\assignments\index.js**

Catches and sets an error if unable to create an assignment successfully

**frontend\src\services\index.js**

Displays the error message that was set by the caught error


defkorean (commented):
Looks good, but is there a case of updating an assignment name to the same name of another assignment? Or would this be a separate issue to look into? Ticket 84 doesn't have specific details as to what the intended purpose is to discourage assignment name duplication, but if it prevents creating assignments with duplicate names I think there should be a case for disabling duplication of name when updating an assignment as well. For example, let's say we create two assignments called a1 and a2 with all the same fields checked (due date, autograder points, etc). Let's say we go into a2's settings and change its name to a1. Currently, it successfully changes the name to a1.

kiat:
braden-carter will do the changes on this PR. and then please check again ...

EthanZhang15 (approved):
the assignment names cannot be the same on creation worked, but renaming an assignment would give an error. This could be just my machine, looks good otherwise.

braden-carter:
Updating the assignment's name now works properly to prevent a duplicate name but allows a valid change.

defkorean (approved):
Thanks!

Issue #38, opened 2023-06-27
Refactor backend


allenjeffreywu:
Combine student and instructor tables to create a user table. This should result in a single user.py in the backend that handles all students and instructor accounts based on role.

The table should have a new role column (identify a user as student, instructor, or even admin)

This is a breaking change, when this ticket is merged please notify if the slack channel

kiat:
We need to have a single user table and user roles like having instructor, student, and teaching assistant.

maximgao:
Looks like this has been implemented

PR #94, opened 2024-07-19
Changes to the Database

Vellankijay:
I've made changes to the structure of the database. The new format centralizes all of the users under one table. Additionally, the tables and API endpoints that are no longer needed have been removed. Further expected change is that the frontend implementation of the Login has been generalized and doesn't depend on the role of the person. These changes shouldn't affect the functioning of the application before the updates.

Changes to come:
- A frontend implementation of the TA's (shouldn't affect the current operations)
- Addition of any user to the course (shouldn't affect the current operations)


defkorean:
If I'm not wrong, the change to the database is a reflection of issue ticket #38 correct? Also, just to clarify, the "changes to come" is referring to the changes that you made through these two commits and are awaiting for review in this pull request?

logankimm:
For TA implementation, wouldn't it be better to
1. For the role, only make it so can be a binary choice between teacher and student
2. Add an option for a TA within the course table
In a scenario where a user is both a student in a course and a TA for another course, then they would have to create two unique accounts. But if we want the user to be able to sign into their UT ID, then that wouldn't be possible with the current implementation

backend/api/create_tables.sql
```diff
@@ -1,15 +1,11 @@
 /* Create Students table */
```
defkorean, line 1: Update comment to show Users

backend/api/create_tables.sql
```diff
@@ -23,15 +19,15 @@ CREATE TABLE courses (
     entryCode varchar(50),
     allowEntryCode boolean DEFAULT FALSE,
     description varchar(100),
-    FOREIGN KEY (instructor_id) REFERENCES instructors (id)
+    FOREIGN KEY (instructor_id) REFERENCES people (id)
```
defkorean, line 22: maybe for consistency's sake, we could change this to users rather than people

backend/api/models.py
```diff
@@ -3,21 +3,15 @@
 from sqlalchemy.types import LargeBinary
 from api import db
 
-class Student(db.Model):
-    __tablename__ = "students"
+class People(db.Model):
```
defkorean, line 6: I believe having Users as the name would be better than just People

docs/CodeAssist_Design_Doc.md
defkorean, line 1: Good job on adding to the design document.

frontend/src/pages/home/logInModal.js
```diff
@@ -1,8 +1,8 @@
 import { Button, Form, Input, Modal, Radio } from "antd";
-import { useContext } from "react";
+import { useState, useContext } from "react";
 import { GlobalContext } from "../../App";
 // import axios from "axios";
-import { instructorLogin, studentLogin } from "../../services/user";
+import {userLogin } from "../../services/user";
```
defkorean, line 5: small style change here, add a space between the curly brace and userLogin

frontend/src/pages/home/logInModal.js
```diff
@@ -15,21 +15,34 @@ export default function LogInModal({ open, onCancel, logIn }) {
...
     try {
 
-      if (isStudent) {
-        res = await studentLogin(restValue);
-      } else {
-        res = await instructorLogin(restValue);
-      }
+      res = await userLogin(restValue);
+      
+      // if (role == 'student'){
```
defkorean, line 26: I'm assuming this code is here for later changes of differentiating roles and doing different things right?

frontend/src/pages/home/logInModal.js
```diff
@@ -15,21 +15,34 @@ export default function LogInModal({ open, onCancel, logIn }) {
...
+      //   updateUserInfo(userInfo);
+      // }
 
       if (res) {
         const userInfo = {
           name: res.data?.name,
           id: res.data?.id,
-          isStudent,
+          isStudent: res.data?.role === 'student',
+          isTA: res.data?.role == 'TA',
```
defkorean, line 45: Pretty sure you need to change the '==' to '==='

frontend/src/pages/home/signUpModal.js
```diff
@@ -12,16 +12,21 @@ import { signUpInstructor, signUpStudent } from "../../services/user";
...
+    const { role, ...restValue } = values;
     let res;
-    if (isStudent) {
-      res = await signUpStudent(restValue);
-    } else {
-      res = await signUpInstructor(restValue);
+    // let taRes;
+    res = await signUpUser(restValue);
+
+    if (role == 'student'){
```
defkorean, line 25: I believe you need to use '===' here as well by convention

defkorean (changes requested):
Biggest things I noticed is style inconsistencies. One of the suggestions I would consider is changing all instances of 'People' to 'User' since I'm noticing the endpoints are using User rather than People. As for functionality, it works well and I haven't run into a bug that crashes the site. Please review these comments and feel free to let me know if any clarification is needed anywhere.

backend/api/models.py
hudsonwhipple: I agree, I think it should say users in each place instead of people

backend/routes/user.py
```diff
@@ -42,217 +42,66 @@ def create_user():
...
     '''
     /get_instructor_by_id gets the student from the database
     Requires from the frontend a JSON containing:
     @param id    the instructor id
     '''
     insid = request.args.get("id")
 
-    instructor = db.session.query(Instructor).filter_by(id=insid)
-    instructor = StudentSchema().dump(instructor, many=True)[0]
+    person = db.session.query(People).filter_by(sis_user_id=insid)
```
hudsonwhipple, line 104: Should it filter by id instead of sis_user_id?

frontend/src/pages/home/logInModal.js
Vellankijay: It was supposed to earlier, but it serves no function right now.

defkorean:
Did you mean to close this pull request? None of your commits has been merged to main

Vellankijay:
> Did you mean to close this pull request? None of your commits has been merged to main

Yes because this current request is very behind main so I am creating a new PR with the updated code

PR #98, opened 2024-08-01
Implemented DuplicateAssignmentModal.js

EthanZhang15:
adds the feature for instructors to duplicate an existing assignment and give it a new name.


kenlu66 (approved):
All the changes look good to me!

PR #99, opened 2024-08-06
Changes to Database

Vellankijay:
I've made changes to the structure of the database. The new format centralizes all of the users under one table. Additionally, the tables and API endpoints that are no longer needed have been removed. Further expected change is that the frontend implementation of the Login has been generalized and doesn't depend on the role of the person. Another change would be the ability for the instructor to freely change the role of the students enrolled to TAs and vice versa. These changes shouldn't affect the functioning of the application before the updates.


defkorean (changes requested):
Looks good to me, but I suggest adding/removing endpoints in the backend_endpoints doc for developers to reference to. I know this was done in the previous pull request, but I don't see those changes here.

logankimm:
Also looks good to me. Just needs update for documentation. I think there will be need to edit/make sure all the existing endpoints work once the database changes, so it might be a good idea to create an issue for that.

defkorean (approved):
We can add documentation to the changed backend endpoints in a later issue ticket. As far as functionality goes, it looks good to merge. One thing to do is to resolve merge conflicts (probably accept all incoming changes from latest commit of this issue ticket in test.py)

EthanZhang15 (approved):
should be good to go besides the endpoints

kenlu66 (approved):
It looks good to me. Update to last commit to avoid conflicts before merge

PR #100, opened 2024-08-08
Ticket 96 - Updated documentation for a streamlined process of setting up coding environment 

defkorean:
Made changes primarily to README.md

Please leave any comments if there needs to be further clarification or added content.


logankimm:
I think on line 42-53 the information of creating the string/link to the database connection should still be there. I know you have it later within the file, but it makes sense intuitively to me that there should be a link connecting to the database, and then the user sets up the database/docker container. Other than that, looks good to me.

kiat (approved):
Thanks!

Issue #95, opened 2024-07-21
Due Date Restrictions for Submissions


kenlu66:
Currently, we do not enforce due date restrictions for assignment submissions. This will allow students to be able to submit or resubmit assignments at any time. We need to implement functionality to restrict submissions based on the due date and late submission policies set by the instructor.

**Expected situation:**
1. No late submissions allowed:
- Students should not be able to submit or resubmit once it has passed the due date if the instructor does not accept late submissions.
- Attempts after the due date will lead to a message indicating that submissions are unsuccessful and no further submissions will be allowed.

2. Late submissions allowed:
- Students will be able to submit or resubmit after the due date.
- Upon submission, students should be clearly informed that their submissions will be counted as late.
- The submission should be marked with a "LATE" sign to indicate the status to both students and the instructor.

**Additional consideration:**
- Give instructors options to either allow or not allow late submissions. Potentially while creating the assignment, and allow them to change that later.
- Consider time zones might affect.

PR #102, opened 2024-10-23
Late submissions

zachli88:
Based on issue ticket #95 

Enforce due dates, but still allow for submissions and resubmissions before a secondary "late due date" if instructor allows for it. Submissions after regular due date but before late due date will be marked as late. Students will be notified that their submissions will be marked late prior to submitting.


kenlu66 (commented):
All looks good to me!

kenlu66 (approved):

gioeloja:
looks good, is this ```logfile``` supposed to be included though?

PR #109, opened 2024-11-03
Ansh/add a user 106

anshramanath:


backend/routes/user.py
```diff
@@ -89,9 +89,9 @@ def get_users():
     # print(result)
     
     # TODO THIS IS A MAJOR SECURITY VULNERABILITY, IT SHOWS PASSWORDS!
-    if len(result) == 0:
+    if not(result):
         return "No user found", 404
-    return jsonify(result)
+    return jsonify(result[0])
```
gioeloja: I would just make sure that there aren't other components that are using this API because it could mess up how they work. If there is, it might make more sense to make this change at the component level

Issue #111, opened 2024-11-06
Prohibit multiple accounts being made under one email


gioeloja:
**Right now, one email can be used to create several accounts. We do not want to allow this.**

```
curl "http://localhost:5001/get_users?email=student2@gmail.com"

[
  {
    "email_address": "student2@gmail.com",
    "id": "283dfd81-37f6-4dab-a503-ffcc7ba44ce9",
    "name": "student2",
    "password": "test",
    "role": "student",
    "sis_user_id": "student2"
  },
  {
    "email_address": "student2@gmail.com",
    "id": "fc312759-353d-49e6-882f-c5f735f9e766",
    "name": "student2",
    "password": "testt",
    "role": "student",
    "sis_user_id": "studentsecond2"
  },
  {
    "email_address": "student2@gmail.com",
    "id": "4a4f83c8-7086-4a0d-9b2a-4ee7de5cda7d",
    "name": "studentthird2",
    "password": "test",
    "role": "student",
    "sis_user_id": "student2222"
  }
]
```

**To do:**
- Edit ```/create_user``` endpoint to check if an account already exists with this email. If there is, return error

gioeloja:
Closed by #113.

PR #113, opened 2024-11-11
Enforcing unique user emails and course entry codes

gioeloja:
Prevent multiple accounts being made with the same email by enforcing unique emails. Similarly, preventing multiple courses from having identical entry codes by enforcing unique entry codes on the SQL table definitions.

Closes out #111.


kenlu66 (approved):
All looks good, a unique email/entry code is required

kiat:
Thanks!

Issue #116, opened 2024-11-13
Network Error when entering an invalid course code


ruolin111:
On the student page, when entering a course entry code that does not exist, a network error pops up. Should show an error message saying that the course does not exist instead of giving a network error. 

Pictures:
![image](https://github.com/user-attachments/assets/199a9746-9241-4df6-9c2d-13a78327dbde)

Network Error:
![image](https://github.com/user-attachments/assets/b91af391-b9b2-4f0f-b3c1-86b28a12c642)

maximgao:
Original network error msg was removed, but I'm seeing 2 error messages pop up: 1st is empty, 2nd says that class doesn't exist.

PR #124, opened 2024-11-18
Fixed bug #116

ruolin111:
Before: Network error when a student inputs an invalid course code when adding a course
Now: An error banner will pop up to inform the user that the course code is invalid.


gioeloja (approved):
Looks good.

kiat:
Thanks!

PR #125, opened 2024-11-18
duplicate Delete Course button removed

kenlu66:
The "Delete Course" without functionality is now removed from course settings


ruolin111 (changes requested):
Revert/remove the changes in the docker-compose.yml. This should not be changed in the main repo.

PR #132, opened 2025-02-04
defined missing variables

manafasif:
Defined the following: 
`const [lateDueDate, setLateDueDate] = useState();
  const [lateAllowed, setLateAllowed] = useState;`
  
 Which appeared to be undefined in the previous commit.


donnesu (approved):
looks good

PR #139, opened 2025-02-11
Fix settings update bug and enabled directly editing course settings (#118)

### LINKED ISSUE #118: Update course bug when updating both basic settings and grading defaults
On the instructor side, on the course setting page, when trying to update any setting under "Basic Settings" and any setting under "Grading Default", the basic settings will not be updated after clicking the "Update Course" button, and the user will be returned to the course dashboard page. However, if none of the settings under "Grading Default" is chosen, changed settings under "Basic Settings" will be correctly updated. 

This may be because the settings under "Grading Default" are not implemented in the backend. 

image

manafasif:
Fixes #118 

Course settings form was previously not updating GlobalContext, but fixed it by calling updateCourseInfo. 

Reimplemented course settings form with antd's "Form.useForm();", which automatically manages form state. This simplifies code for the form.

Enabled users to directly edit the previous values for each course setting, instead of having to retype them. This is accomplished by using the form state to load the previous course settings as "values" for each input [see line 48], rather than as "previews" (as it was previously done).


maximgao:
- "Allow students to enroll via course entry code" checkbox does not stay checked upon saving
- Looking through `courseSettings/index.js`, I see that some functions can be turned async. Unsure if this is out of scope of the ticket though
- Otherwise, seems to work fine

manafasif:
Just pushed a fix for the checkbox @maximgao

maximgao:
LGTM

maximgao (approved):

Issue #148, opened 2025-02-20
Assignment Submission Autograding - Implement Asynchronous Job Processing with Status Monitoring


manafasif:
> Implement asynchronous grading. Submitting assignment should return an OK/Fail response, triggering a "grading" job and closing the upload window. 
> 
> The job should be handled by a separate thread (from main thread), which updates some in-memory variables relating to the status/progress of the job.
> 
> Frontend can continously poll a "status" endpoint to read the status of the job until it finishes. 
> 
> This is essential, as current system will fail to handle a long-running autograder, or multiple submissions at once, which will slow the system. 

 _Originally posted by @manafasif in [#147](https://github.com/kiat/codeAssist/issues/147#issuecomment-2670489091)_

manafasif:
See this example implementation to get started:

### **System Design**

- **Job Queue**: Use an in-memory job queue (e.g., a Python `queue.Queue` or a simple dictionary in memory).
- **Job Worker**: Runs in a separate thread to simulate asynchronous processing.
- **Job Status Store**: Use an in-memory dictionary or a simple database (like SQLite) to store job statuses.

---

### **Backend Implementation**

#### 1. **Job Queue & Status Store**
We will use Python’s `threading` and `queue` modules to simulate job processing in the background and track the job status.

**`app.py`**:
```python
from flask import Flask, request, jsonify
import uuid
import threading
import time
from queue import Queue

app = Flask(__name__)

# In-memory structures
job_queue = Queue()  # To queue jobs for processing
job_status = {}  # Dictionary to store job statuses

# Job Worker Thread to process jobs asynchronously
def job_worker():
    while True:
        # Get the next job from the queue
        job_id, data = job_queue.get()
        if job_id is None:
            break

        # Simulate job processing with progress updates
        for i in range(1, 11):  # Simulate 10 steps
            time.sleep(1)  # Simulate work (1 second per step)
            job_status[job_id]['progress'] = i * 10
        job_status[job_id]['status'] = 'completed'

@app.route('/submit-job', methods=['POST'])
def submit_job():
    # Generate a unique Job ID
    job_id = str(uuid.uuid4())
    
    # Save initial job status
    job_status[job_id] = {'status': 'pending', 'progress': 0}
    
    # Add job to queue
    job_queue.put((job_id, request.json))
    
    # Return immediate response with Job ID
    return jsonify({"job_id": job_id}), 202

@app.route('/job-status/<job_id>', methods=['GET'])
[...]

PR #154, opened 2025-02-21
Implemented lighter Docker image and optimized setup commands

manafasif:
Changed the autograder setup to use a lighter docker image (python:3.9-slim) and chained together the run commands for efficiency.

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -212,10 +212,12 @@ def upload_assignment_autograder():
     filepath = os.path.join(assignment_dir, filename)
     file.save(filepath)
 
-    # query teh db and find out if this assingnet in the db alreayd has an associated container or not
+    # Query the database for the assignment
     assignment = db.session.query(Assignment).filter_by(id=assignment_id).first()
     if not assignment:
         return jsonify({"error": "Assignment not found"}), 404
+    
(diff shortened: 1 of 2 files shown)


backend/routes/submission.py
```diff
@@ -288,17 +289,10 @@ def upload_assignment_autograder():
     os.chdir(current_dir)
     subprocess.run(f"docker stop {container_name}".split(), capture_output=True)
 
+    # Update the assignment record with the new container ID
     assignment.container_id = container_name
-    # assignment.container_id = container_id
     db.session.commit()
 
-    # Extract the contents of the uploaded ZIP file
```
maximgao, line 295: What was the purpose of these lines? Just to detect if the zip file is valid?

manafasif: Yeah, seems like it just tries to extract the zip file

The extraction already happens in the container, and these lines were commented out, so I decided to delete them.

gioeloja:
@manafasif Changes look good, were there any noticeable speed changes in submissions or the backend tests?

gioeloja (approved):

manafasif:
@gioeloja maybe a little bit, it wasn't too much

I think the biggest difference will come once we implement #148

Issue #126, opened 2024-11-20
Implementing autograder timeout


gioeloja:
To prevent long/infinite runtimes of student submissions, the autograder must have a timeout feature. This parameter should be included in the ```/upload_assignment_autograder``` API.

To do:
- Have docker image timeout if a submission runs too long, should return an error specifying that a timeout occurred.
- Add timeout setting in frontend (autograder configuration). 
- On student side, if a submission results in a timeout, this error should be displayed.
- Implement tests for this

gioeloja:
Gradescope: 

![Image](https://github.com/user-attachments/assets/3f593412-eb88-46bc-ba08-7fb8edb63675)

PR #157, opened 2025-02-25
Created autograder timeout feature #126

gioeloja:
**Backend changes:**

- Added autograder_timeout field to submission DB entries
- In the /upload_submission endpoint, added timeout option so that docker process stops running if this time is exceeded
- Added new timeout error for error code 405
- Refactored /upload_submission endpoint to use the new error system

**Frontend changes:**
- On teachers view, added new autograder timeout option in the autograder configuration screen.
image

- On student's view, if their submission times out, this error is displayed.
image

### DIFF
--- a/backend/api/create_tables.sql
+++ b/backend/api/create_tables.sql
@@ -62,6 +62,7 @@ CREATE TABLE assignments (
     autograder_file bytea,
     -- new container_id column
     container_id varchar(),
+    autograder_timeout integer DEFAULT 300,
     FOREIGN KEY (course_id) REFERENCES courses (id)
 );
 
--- a/backend/api/models.py
+++ b/backend/api/models.py
@@ -48,6 +48,7 @@ class Assignment(db.Model):
     published_date = db.Column(TIMESTAMP, nullable=True)
     autograder_file = db.Column(LargeBinary, nullable=True)
     container_id = db.Column(db.String)
+    autograder_timeout = db.Column(db.Integer, default=300)
 
 class Submission(db.Model):
     __tablename__ = "submissions"
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -134,7 +134,7 @@ def upload_submission():
             timeout=autograder_timeout)
     except subprocess.TimeoutExpired:
         os.chdir(current_dir)
-        raise ServerTimeoutError("Autograder execution timed out")
+        raise ServerTimeoutError("Autograder execution timed out. Submitted program took too long to run.")
 
     if exec_proc.returncode != 0:
         os.chdir(current_dir)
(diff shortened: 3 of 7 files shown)


gioeloja:
@maximgao assigned you to look at if I did the error/axios stuff correctly

maximgao (changes requested):
Axios interceptor and Flask handler stuff looks correct

## Changes
- `uploadSubmission` service function already exists in `/services/assignmentResult.js`, but currently not being used
- `InternalProcessingError` messages are shown directly to the user, so exception details should probably be abstracted away for the user and can be logged to console or something if needed. Specifically, I see docker errors and the file delete failure sharing `e` itself.
- ~~Please change `ServerTimeoutError` error code 405 to 504 (see [this](https://stackoverflow.com/questions/58875992/what-is-the-correct-http-response-code-to-send-if-server-times-out-to-process-a))~~
## Minor Notes / Out of scope
- Another small thing is the formatting of the error message. Timeout error message currently uses 2 periods, I've been writing user error messages as 1 phrase and without punctuation and defined it this way in guidelines documentation.  Minor thing though tbh, but could help consistency in future. Guidelines can be changed if you think this format is better.
- Regarding commands for spinning up the docker containers, potential code execution vulnerabilities? Probably warrants project-wide investigation for input sanitization in a later ticket.

- Regarding frontend service organization, I think that the Axios service functions should be grouped together based on which resource the backend endpoint is actually a part of. For example, `/upload_submission` is a owned by `submission` resource, so `uploadSubmission()` should be in `/services/submission.js` file. This will require a refactor of all the existing methods as well, but should help a lot with organization. But, looks like is out of scope for this ticket.

gioeloja:
> ## Changes
> * `uploadSubmission` service function already exists in `/services/assignmentResult.js`, but currently not being used
> * `InternalProcessingError` messages are shown directly to the user, so exception details should probably be abstracted away for the user and can be logged to console or something if needed. Specifically, I see docker errors and the file delete failure sharing `e` itself.
> * Please change `ServerTimeoutError` error code 405 to 504 (see [this](https://stackoverflow.com/questions/58875992/what-is-the-correct-http-response-code-to-send-if-server-times-out-to-process-a))
> 

Thanks for the review, didn't see the already existing ```uploadSubmission```. For the error code, ```ServerTimeoutError``` is already 504. Did you want to change this?

maximgao:
Oops got confused, no that's fine.

gioeloja:
@maximgao I put in the changes... made it so the specific error is printed and just "Failed to grade submission" is shown to the user.

gioeloja:
@maximgao removing the submission service file wasn't included in my prev commit for some reason but I've included it now.

kiat:
Thank you!

PR #164, opened 2025-02-26
refactored /services to match /routes, fixed init_db.py

### LINKED ISSUE #158: Refactor frontend services organization
**Problem Statement**
The files in  `src/services/` directory should be reorganized such that the methods are contained within a file named after their corresponding owner resource. For example, `uploadSubmission()` method which calls `/upload_submission` endpoint should be in a file named `submission.js`, as the endpoint is part of the `submission` resource, or `routes/submission.py`.

Careful of frontend imports that get messed up.

Image

**Proposed Solution**
- create new files mirroring backend's `/routes` directory
- move existing service methods to corresponding resource file
- fix dependency imports

maximgao:
- refactored `/routes` directory service methods
- fixed `init_db.py` to work with factory pattern

closes #158


EthanZhang15 (approved):
fixed the init db

Issue #151, opened 2025-02-20
Fix Course Dashboard Formatting


maximgao:
**Problem Description**
- Course dashboard formatting is currently not correct for both student and instructor displays.

**Proposed Solution**
- The course cards should be grouped in sections by semester, filling rows from left to right
- Each semester section should start with a header stating the semester name (Fall 2025) and restart from the leftmost position
- Headers should probably be in reverse chronological order from top to bottom

- I think the "Add Course" button should stay at the top left within it's own section

<img width="1152" alt="Image" src="https://github.com/user-attachments/assets/b729d966-14b5-45d5-963a-7cd506ffc442" />

For reference, this is what Gradescope actually looks like:

![Image](https://github.com/user-attachments/assets/c600cc92-f04a-4918-94e9-a69b70dad506)

PR #172, opened 2025-03-04
clean up course block styles

donnesu:
Edited styling of course blocks, and flexbox grouped by semester; not yet ordered

Before:
Screenshot 2025-03-04 at 3 54 35 PM

After:
Screenshot 2025-03-04 at 3 46 07 PM

#151


maximgao:
~I'm seeing it show up like this, going column-by-column instead of row-by-row.~

Also PR is necessary if sorting is not implemented? Might as well close the entire ticket with 1 PR.

My fault, didn't pull newest changes on the branch. Now what I see:

image

donnesu:
hmmm, not sure why it's not showing up

I'll look into fix and implement sort.

PR #173, opened 2025-03-04
removed docker_client dependency in submission.py

maximgao:
- removed `import docker_client` dependency which required docker client to be open when running tests
- docker client no longer needs to be open when running integration tests

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -1,5 +1,4 @@
 import uuid
-import docker_client
 import json
 import sys
 from werkzeug.utils import secure_filename


donnesu:
looks good

kenlu66 (approved):
LGTM
If using Docker SDK only requires: import docker, right?

maximgao:
> LGTM If using Docker SDK only requires: import docker, right?

Looks like it, although we can revert back for that ticket if not the case.
Just doing this so that Docker Desktop doesn't have to be open when running tests.

PR #176, opened 2025-03-07
useEffect loop fixed on Instructor Course Dashboard

### LINKED ISSUE #152: Constant GET calls in instructor view specific course dashboard
**Problem Description**
- When on the dashboard for a specific course as an instructor, HTTP GET requests are constantly called
- Specific endpoints are shown in the screenshot

**Proposed Solution**
- Investigate root cause and fix
- I suspect some webhook issue that's constantly refreshing?

Image

manafasif:
Fixes #152 

Found a dependency loop in the useEffect on the instructor's course dashboard. Changed the dependencies to resolve it.

After fixing it, found that the dashboard wasn't updating as expected when course settings were changed. Added lines in the courseSettings page to update the local copy of course settings once they are updated on the server side.

### DIFF
--- a/frontend/src/pages/instructor/courseSettings/index.js
+++ b/frontend/src/pages/instructor/courseSettings/index.js
@@ -29,6 +29,9 @@ export default () => {
   // Used to manage form state
   const [form] = Form.useForm();
 
+  const { courseInfo, updateCourseInfo } = useContext(GlobalContext);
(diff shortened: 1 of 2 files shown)


maximgao (approved):
LGTM

kenlu66:
LGTM as well

PR #187, opened 2025-04-03
Tests for submission

aidan-liu:
wrote tests for testing routes, retrieval, deletion, submission details, active submission.


backend/test/unit/test_submission.py
```diff
@@ -0,0 +1,284 @@
...
+#     assert response.status_code == 404
+#     data = response.get_json()
+#     assert data["message"] == "No submission found to delete"
+
+
+# # FAILED TEST
+# def test_upload_assignment_autograder_missing_file(client):
+#     """Test /upload_assignment_autograder returns error when file is missing."""
+#     response = client.post("/upload_assignment_autograder", data={})
+#     assert response.status_code == 400 or response.status_code == 500  # depending on error handling
```
maximgao: What's the reason why these are commented out? Just that they're failing? Or missing feature?

aidan-liu: They are failing, just unimplemented features for now

maximgao: No I mean why are they failing, is it because there's missing error-handling or the flows are incorrect? Or maybe the tests are just written wrong.

aidan-liu: fixed the errors, mainly caused by the submission.py route, had missing status codes 
ex: return jsonify({"error": "not sufficient details"}) 
changed to
return jsonify({"error": "not sufficient details"}), 400

and also updated the tests so they all pass now

backend/routes/submission.py
```diff
@@ -210,7 +210,7 @@ def upload_submission():
 def upload_assignment_autograder():
     # Validate input file and parameters
     if "file" not in request.files:
-        return BadRequestError("No file part")
+        return jsonify({"error": "No file part"}), 400
```
maximgao: This should be reverted, we want to use the unified error handling classes.
Although, the line should be:
`raise BadRequestError("No file part")`

Looks like slipped by PR review.

backend/routes/submission.py
```diff
@@ -397,7 +397,7 @@ def delete_submission():
         return jsonify({"error": "Missing submission_id"}), 400
 
     # Query for the submission to be deleted by submission_id
-    submission_to_delete = Submission.query.get(submission_id)
+    submission_to_delete = db.session.get(Submission, submission_id)
```
maximgao, line 400: good catch, looks like query on specific db resource is deprecated and all instances should be changed to `db.session` in the future

backend/routes/submission.py
```diff
@@ -443,7 +443,7 @@ def get_active_submission():
     assignment = request.args.get("assignment_id")
 
     if not assignment or not student:
-        return jsonify({"error": "not sufficient details"})
+        return jsonify({"error": "not sufficient details"}), 400
```
maximgao: Should probably be changed to `raise Error` error-handling format.

maximgao: Also, fyi the repo uses `"message":"message goes here"` instead of `"error"` for error messages. Not sure why this design decision was made.

maximgao (changes requested):

backend/routes/submission.py
aidan-liu: ok fixed to 
raise BadRequestError("No file part")

backend/routes/submission.py
aidan-liu: ok changed to the raise error format

maximgao (approved):
Tests LGTM

Note that coverage of `submission.py` is only at 46%.
Also, consider adding more call count checks in the future, but these can be added later.

Issue #185, opened 2025-04-01
Handle Null Late Due Dates on Dashboard


manafasif:
Currently, if a late due date isn't specified for an assignment, it shows up like this on the dashboard:

![Image](https://github.com/user-attachments/assets/f0612862-4ac1-440d-9fcf-41ef44f181b8)

Add a check for a null late due date on the dashboard, and make it say something like "None", instead of "Invalid Date"

Issue #184, opened 2025-04-01
Check Autograder File Type Before Uploading


manafasif:
While uploading the autograder, I accidentally selected a .py file and it caused this error:

![Image](https://github.com/user-attachments/assets/1b6bea5f-8888-41ff-8064-d789734f4c17)

Implement a check on the frontend to make sure that the user is uploading a .zip, and deny the form submission gracefully

PR #191, opened 2025-04-08
Handling null dates on dashboard + checking autograder file type

gioeloja:
#184 and #185 

For ```null``` dates, student assignment dashboard now displays "None" instead of "INVALID DATE".
image

When upgrading autograder files, the frontend verifies that the uploaded file is a .zip now and displays an error if otherwise.
image


maximgao (approved):
LGTM

kiat:
Thanks!

PR #193, opened 2025-04-09
Dashboard Buttons Issue Fix

manafasif:
Rolled back some changes that were made to the course dashboard to resolve a bug we found where new accounts would not have their dashboards loading properly

https://github.com/kiat/codeAssist/commits/main/frontend/src/pages/dashboard/index.js

I'm not sure if this is a complete fix, it will need to be tested more thoroughly

In the meantime, I've pushed it so its no longer a blocker


maximgao:
Should be resolved by #199

PR #195, opened 2025-04-09
fixed bad endpoint GET /get_assignment

manafasif:
endpoint returns a query object, not an actual assignment

this results in the assignment settings page not working properly


aidan-liu (approved):

kiat:
Thanks!

PR #198, opened 2025-04-09
Implemented better default values for the assignment creation page

manafasif:
Set default autograder points to 100
Set default release date to now 
Set default due date to 1 week from now

makes it easier, quicker, and more intuitive to create an assignment


gioeloja (approved):
Looks good

PR #199, opened 2025-04-10
empty dashboard bug fix

maximgao:
Issue was that `+ Add a course` button would not render if there were no classes added.

Fixed with fall-back clause.


kenlu66 (approved):
LGTM

kiat:
Thanks!

PR #213, opened 2025-04-23
Password Encryption Implemented

manafasif:
Encrypts API keys and passwords in the database if encryption keys are defined in .env:
`API_SECRET_KEY=...
PASSWORD_SALT=...`

These can be generated automatically by running backend/init_encryption_keys.py

Test without encryption enabled, then enable it and make sure everything still works. 

Note that if encryption keys change or you go from unencrypted to encrypted (or vice versa), the old logins will stop working since they were created with a different key, so clear the database when testing.

### DIFF
--- a/backend/ai_integration.py
+++ b/backend/ai_integration.py
@@ -19,6 +19,7 @@ import threading
 from sqlalchemy.orm import aliased
 import json
 
+from util.encryption_utils import decrypt_api_key
 
 
 DEFAULT_MODEL="gpt-4-turbo"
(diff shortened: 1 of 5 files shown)


kiat:
Please check the failing tests!

manafasif:
> Please check the failing tests!

Fixed!

gioeloja (approved):
looks good, probably just need to update documentation (main readme?) with info about these ```.env``` variables.

kiat:
Thanks!

Issue #214, opened 2025-04-25
```upload_submission``` route: timeout submission handling bug


gioeloja:
Currently, if a submission times out, we are only displaying a popup error to the user. We do not upload an actual submission to the database. Because of this, a frontend error occurs if this submission is the first submission for this assignment. 

To fix this, we should upload a ```Submission``` database entry for a submission that times out. This way, our frontend can load the submission.

Additionally, the docker container should be cleaned up if a timeout occurs (files should be removed).

PR #215, opened 2025-04-25
Fixing submission timeout handling 

gioeloja:
#214

Changes:
- When a submission times out, a ```Submission``` database entry is made.
- Docker container files are cleaned up if a timeout occurs.
- Loader added for first submission upload modal (the upload modal that pops up when you click on the assignment from dashboard). This modal has also been refactored to use updated axios handlers.
- New ```SubmissionTimeoutError``` that is a subclass of ```ServerTimeoutError``` but with ```submission_id```. This is necessary for navigating to the results page after a submission times out.

image


kiat:
Thanks!

PR #223, opened 2025-05-21
Fix/handle submit without autograder #196

### LINKED ISSUE #196: Handle Submitting without Autograder Configured
Right now, if the student tries to submit before the instructor configures an autograder, they get this error

Image

and a bunch of errors on the server

Image

submission should still work without an autograder, it should just not try to run the autograder

shreyavarma99:
Fixes #196
- Show submitted code even when autograder is missing
- Prevent frontend crash on missing results
- Display "Ungraded" instead of score if no grade
- Toggle "Graded" checkmark based on actual score presence
image

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -116,7 +116,7 @@ def upload_submission():
             execution_time=0.0,
             submitted_at=datetime.now(),
             active=True,
-            completed=False,
+            completed=True,
             submission_number=submission_count + 1,
             ai_feedback=None
         )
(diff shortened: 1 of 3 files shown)


yangshao05 (commented):
When a student submits an assignment that has no autograder, the submission is saved, but it does not appear on the instructor side. It seems the issue is caused by this part of the code: 
`old = db.session.query(Submission).filter_by(...).update({'active': False})` 
This doesn’t deactivate previous submissions.

yangshao05 (approved):
Update: re-tested the changes and the student panel correctly updates to “completed.” The issue I raised earlier is resolved. Thanks for addressing it! Sorry for the confusion earlier!

PR #224, opened 2025-05-22
Fix: Prevent using old file during autograder upload

### LINKED ISSUE #192: ```upload_assignment_autograder``` route: updating autograder file desync bug
When uploading an autograder file, the first upload will work as expected. If followed up by another upload of a different autograder, the backend throws an error because it is attempting to work with the file from the first upload. 

Frontend seems to be correctly passing in the correct and current file, so I suspect the problem is in the docker file cleanup in the ```upload_assignment_autograder``` code.

yangshao05:
Fixes #192 

- Fixed a bug where the backend re-used the previously uploaded `.zip` file when a new autograder was uploaded
- Ensured Docker container cleanup before processing new uploads
- Added diagnostic logging `(ls -l /autograder/)` to verify file presence inside the container

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -268,6 +268,12 @@ def upload_assignment_autograder():
             container = docker_client.containers.get(container_id)
             container.start()
 
+            # Force cleanup of all files in /autograder before uploading new ones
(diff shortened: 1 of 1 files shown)


kiat:
Thanks!

Issue #217, opened 2025-04-25
Test Autograder


donnesu:
Currently Test Autograder is not implemented with any functionality.

Tooltip is showing HTML error response (from failed POST) as a file preview. 
Most likely coming from Ant Design’s <Upload.Dragger> component, which tries to preview files
<img width="539" alt="Image" src="https://github.com/user-attachments/assets/17bbb489-b3d0-4b30-b9e0-e5035e4f0ebc" />

PR #231, opened 2025-06-05
Fix/test autograder 217

shreyavarma99:
#217 Added functionality to "test autograder" button:
  - Accepts a submission file and autograder zip file.
  - Builds a temporary Docker container to run the autograder.
  - Returns the results of the grading run without saving any data to the database.
  - Cleans up all temporary files and containers after the run.

Note: Test autograder builds a Docker image for each test run, which may lead to slow response times; we may need to optimize in the future

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -646,5 +646,9 @@ def test_autograder_submission():
         # Remove temp dir
         shutil.rmtree(temp_dir, ignore_errors=True)
 
-    return jsonify({"message": "Dry run successful", "results": result_json}), 200
-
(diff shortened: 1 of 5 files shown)


yangshao05 (approved):
Looks good to me!

aidan-liu (commented):
I'm not sure how consistent it is to change to all " from ', some files use all " and others use ', although there aren't any functional difference, i looked back at some other files like addForm in dashboard and they all use ' for js elements, and " for strings

aidan-liu (approved):

PR #244, opened 2025-06-24
User error handling

WajeehJ:
finished adding more user error handling 

get_user_by_id()
-make sure user_id provided is a proper uuid
-make sure that a user_id is provided in the body of the request
-throw 404 if not found in database

delete_user()
-make sure that user_id is provided in the body of the request
-make sure user_id provided is a proper uuid
-throw 404 if not found in database
-add try and rollbacks for db commits

update_account()
-make sure that the user_id is provided in the body of the request
-make sure user_id provided is a proper uuid
-throw 404 if not found in database
-add try and rollbacks for db commits

-updated some tests that had mistakes in them eg trying to invoke update_user with eid instead of the actual user id
-added unit testing to test these cases as well

### DIFF
--- a/backend/routes/user.py
+++ b/backend/routes/user.py
@@ -131,6 +131,10 @@ def get_user_by_id():
     '''
     insid = request.args.get("id")
 
+    if not insid:
+        raise BadRequestError("Missing user id")
+
+
     instructor = db.session.query(User).filter_by(id=insid)
     instructor = UserSchema().dump(instructor, many=True)[0]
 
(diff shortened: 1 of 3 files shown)


backend/routes/user.py
```diff
@@ -129,21 +129,52 @@ def get_user_by_id():
...
+    except (ValueError, TypeError):
+        raise BadRequestError("Invalid user id")
+
+
+
+    instructor_obj = db.session.query(User).filter_by(id=insid).first() 
+    if not instructor_obj: 
+        raise NotFoundError("User does not exist")
+    
+    instructor = UserSchema().dump(instructor_obj)
```
yangshao05, line 148: Rename `instructor` to `user` since this route is fetching data for general users
```suggestion
    user_obj = db.session.query(User).filter_by(id=insid).first() 
    if not user_obj: 
        raise NotFoundError("User does not exist")
    
    user = UserSchema().dump(instructor_obj)
```

backend/routes/user.py
```diff
@@ -129,21 +129,52 @@ def get_user_by_id():
...
 
-    instructor = db.session.query(User).filter_by(id=insid)
-    instructor = UserSchema().dump(instructor, many=True)[0]
 
     return jsonify(instructor)
```
yangshao05, line 151: Rename to `user`
```suggestion
    return jsonify(user)
```

yangshao05 (approved):
Looks good to me. The only improvement I noticed is that `get_user_by_id` fetches data for users, not just instructors. It would be clearer to update the variable names from `instructor` to `user`, and also the comment for this route needs to be revised.

kiat:
Thanks!

PR #250, opened 2025-07-01
Edit Grade Erroring Out

makyra:
- `set_reviewed` only called when regrade request is present -> no more error messages
- Edited score is now correctly reflected in the grade section of the results page immediately after saving

### DIFF
--- a/backend/routes/regrade_request.py
+++ b/backend/routes/regrade_request.py
@@ -151,11 +151,17 @@ def get_instructor_regrade_requests():
 
     return jsonify(result), 200
 
-@regrade_request.route('/set_reviewed', methods=['POST'])
+@regrade_request.route('/set_reviewed', methods=['POST', 'OPTIONS'])
 @cross_origin()
 def set_reviewed():
+    if request.method == 'OPTIONS':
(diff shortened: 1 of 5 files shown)


yangshao05 (commented):
Looks good, it fixes the bug.

Since no new dependencies were added, we should keep `package-lock.json` and `package.json` unchanged.

shreyavarma99 (changes requested):
Looks good overall! I agree with Yang that the cors and flask dependencies don’t seem necessary in the frontend package.json. Since they’re typically backend packages, adding them here might cause unnecessary bloat or even conflicts during builds.

makyra:
I have just reverted `package.json` and `package-lock.json` and pushed the changes. Now, only the files related to the bug are edited. Thank you for catching that!

kiat:
Thanks!

BenjiAnzaldua (approved):

yangshao05 (approved):

shreyavarma99 (approved):

kiat:
Excellent, Thanks!

Issue #251, opened 2025-07-03
Race condition in backend routes


shreyavarma99:
Bug exposed by upload_assignments stress test. When multiple assignments are submitted simultaneously, system crashes and submissions are corrupted. The issue is most likely due to the fact that routes like upload_submission write to the same results file and use the same docker container.

Fixes we need to implement:
- Spin up a separate, temporary Docker container for each submission to isolate the runtime environment and avoid shared state conflicts.

- Make all file paths (submission directories, results files, temp folders) submission-specific, e.g., include submission_id or a UUID in the directory structure:
        `/runs/<assignment_id>/<submission_id>/submission/`
        `/runs/<assignment_id>/<submission_id>/results/`

- Clean up containers and temp directories after grading to avoid resource leaks.

PR #253, opened 2025-07-10
fixed race condition in upload_submission route

shreyavarma99:
#251  removes the use of assignment.container_id and fully transitions our submission grading system to use temporary containers per submission. Also, ensures that files (eg. results.JSON) are not shared amongst submissions. This prevents race conditions when multiple students submit code at the same time. Should now pass upload_assignments.py stress test.

NOTE: this request includes changes to the database schema. It will error out unless you modify as shown below.

connect to database container:
`docker exec -it postgres_container psql -U postgres -d codeassist
`

add column:
```
ALTER TABLE assignment DROP COLUMN container_id;
ALTER TABLE assignment ADD COLUMN autograder_image_name VARCHAR;
```

to double check that you added the column:
`\d assignments`


BenjiAnzaldua (approved):
Works on my end! Just a small note, when modifying the database schema, the following commands should use assignments instead of assignment to match the actual table name:

ALTER TABLE assignments DROP COLUMN container_id;
ALTER TABLE assignments ADD COLUMN autograder_image_name VARCHAR,

kiat:
Thanks!

PR #261, opened 2025-07-23
Admin Dashboard 03: Manage Course

AzaanWaseem:
Added features to the Admin Dashboard Courses page:

- Clicking "Manage Course" now shows a list of enrolled students.

- Admins can remove students from a course.

- Admins can add students using their EID.

### DIFF
--- a/frontend/src/pages/admin/courses/manage.js
+++ b/frontend/src/pages/admin/courses/manage.js
@@ -345,45 +345,7 @@ export default function AdminCourseManage() {
       >
         <Form
           layout="vertical"
-          onFinish={async (values) => {
-            try {
-              // You need the student's ID, so you may want to search by EID or email
-              // For this example, let's assume you enroll by EID:
-              const res = await fetch(
-                `${process.env.REACT_APP_API_URL}/get_user_by_eid?eid=${values.eid}`
-              );
-              if (!res.ok) throw new Error("Student not found");
-              const student = await res.json();
-
-              // Enroll the student
-              const enrollRes = await fetch(
-                `${process.env.REACT_APP_API_URL}/create_enrollment`,
-                {
-                  method: "POST",
-                  headers: { "Content-Type": "application/json" },
-                  body: JSON.stringify({
-                    student_id: student.id,
-                    course_id: courseId,
-                    role: "student", // <-- add this line
-                  }),
-                }
-              );
-              if (!enrollRes.ok) throw new Error("Failed to enroll student");
-              message.success("Student enrolled!");
-              setShowAddStudent(false);
-
-              // Refresh enrolled students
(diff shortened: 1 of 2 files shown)


frontend/src/pages/admin/courses/manage.js
```diff
@@ -211,9 +318,87 @@ export default function AdminCourseManage() {
...
+                `${process.env.REACT_APP_API_URL}/get_course_enrollment?course_id=${courseId}`
+              );
+              if (studentsRes.ok) {
+                const studentsData = await studentsRes.json();
+                setEnrolledStudents(studentsData);
+              }
+            } catch (err) {
+              message.error("Failed to enroll student.");
+            }
+          }}
```
yangshao05: Since `handleEnrollStudent` is already defined ahead, they can be removed and replaced with a call to that function.
````suggestion
onFinish={handleEnrollStudent}
````

**Review by yangshao05 — CHANGES_REQUESTED** [review]
Looks good overall. The issue I noticed:

- In `manage.js`, `handleEnrollStudent` is already defined, so it can be called directly in the Enroll Student modal.
- The studentColumns table does not display the EID, likely because the backend route `get_course_enrollment` does not include the sis_user_id field in its response.

File: frontend/src/pages/admin/courses/manage.js [hunk 2]
```diff
@@ -25,65 +27,80 @@ export default function AdminCourseManage() {
...
+
         const instRes = await fetch(
           `${process.env.REACT_APP_API_URL}/get_all_instructors`
         );
         const instructorsData = await instRes.json();
         setInstructors(instructorsData);
+
+        // Fetch students enrolled in this course
+        const studentsRes = await fetch(
+          `${process.env.REACT_APP_API_URL}/get_course_enrollment?course_id=${courseId}`
```
yangshao05, line 73: The `get_course_enrollment` endpoint does not return the student’s EID, so the studentColumns cannot display EID properly. Suggest to replace
```
    students = db.session.query(
        User.name, User.email_address, User.id, User.role
    ).filter(User.id.in_(list_of_students))
```
 with 
```
    students = db.session.query(
        User.name, User.email_address, User.id, User.role, User.sis_user_id
    ).filter(User.id.in_(list_of_students))
``` 
in `course.py/get_course_enrollment`

AzaanWaseem: Thank you for the feedback!

makyra (approved):
Everything works as expected for me! I see that the studentColumns table now display the EID, and that I can add and delete students and instructors using the admin dashboard.

yangshao05 (approved):
Looks good to me!

kiat:
Thanks!

PR #262, opened 2025-07-29
Fixed Migration Issues

makyra:
While trying out the email verification function (which we decided not to go forward with), I discovered issues in previous migrations that were preventing successful migrations. Specifically:
- "input_data" of relation "test_cases" already exists
- relation "submission_submitters" already exists

This PR addresses these issues to ensure smooth migration execution.


kiat:
Please review this!

yangshao05 (approved):
Looks good. This fixes the conflicts related to `submission_submitters` and `input_data`, but on my end it still has conflicts with `submissions` table and `file_name` column. The error message is like:
```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateColumn) column "file_name" of relation "submissions" already exists
```
It might be related to how the database is created on different machines. I think it’s fine to go ahead and merge, and for any other issues we can solve them individually.

makyra:
I see. I have added another check that makes sure the `file_name` column is not added again if it already exists.

May I ask you to try pulling and seeing if that fixes the issue? You should pull and run
`docker compose exec backend flask db upgrade`

Thank you!

yangshao05:
Thanks for the new commit. I checked and found that the duplicate column errors occurred because the database had already applied the upgrade previously. I tested by downgrading to the previous version and then running the upgrade again, and everything works well now.

WajeehJ:
I downgraded to 45b5cf6cc787 and then run docker compose exec backend flask db upgrade and then get this issue: sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateColumn) column "submission_number" of relation "submissions" already exists

am i testing this correctly?

makyra:
Hello, thank you so much to everyone who tested this. I realized that the checks I included to fix the migration issues here were just superficial band-aids, not a real fix.  

The actual issue is that our base tables for the Models were never created in the migration history before being altered. When I tried running our very first migration (45b5cf6cc787) on a fresh DB, it immediately crashed with:
`relation "test_cases" does not exist`
This happens because that migration tries to alter `test_cases` without it ever being created first. Without a “create tables” step, a fresh database can never successfully upgrade to head.

The correct fix will be to add a proper initial migration that creates the base tables, then let the existing migrations alter them. However, since the User table is being renamed/adjusted in another PR, it makes sense to wait until that PR merges to main before rebuilding the migration chain.

The code in this branch is not useful in its current form, so I will close this PR and work on a proper migration chain fix once the User table changes are merged.

PR #263, opened 2025-07-29
Fixed Salting Issue in Encryption

BenjiAnzaldua:
Viewing the password encryption code, I noticed that the hashing logic was using a static salt from the .env file, which isn’t secure. If the SALT flag was not met then it would store the password in plaintext. This is a hazard as anyones account could have been victim to a data breach.

This PR replaces that with a proper randomly generated salt for each password using os.urandom(16), and stores the salt+hash together as a base64-encoded string. However I did have to set PASSWORD_SALT to test because I had some trouble importing the `dotenv package.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python-envs.defaultEnvManager": "ms-python.python:system",
+    "python-envs.pythonProjects": []
+}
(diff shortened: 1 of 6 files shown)


BenjiAnzaldua:
Do not approve yet until I see what caused that Coverage Test failure. It was passing on my machine so I just have to go and double check that.

yangshao05 (approved):
Looks good to me!

makyra (approved):
Looks good to me as well!

WajeehJ (approved):
Looks good!

kiat:
Thank you!

PR #265, opened 2025-08-05
Switch Admin Dashboard Login to Google Sign-In Only

krishaaapatel:
Removed the email/password-based login and signup modals in favor of Google Sign-In only, simplifying the login flow and eliminating the need to handle email verification ourselves.


kiat:
Let us postpone it!

PR #266, opened 2025-08-05
AI testgen feature

shreyavarma99:
This PR introduces an AI-powered test generation feature for instructors using CodeAssist. It leverages a Retrieval-Augmented Generation (RAG) pipeline powered by the mistral-instruct model served via Ollama in a Docker container.

The testgen microservice is included in docker-compose and:
    - Starts automatically on app launch
    - Listens on port 5002
    - Exposes a POST /generate-test endpoint

**Workflow:**
- Accepts instructor prompts via a frontend modal
- Retrieves similar assignment examples using RAG (retrieve_similar_examples)
- Prompt-engineers a message combining:
    - Example spec
    - Example solution
    - Example test code
    - New instructor prompt
- Sends the combined prompt to the Mistral model (/api/generate)
- Returns generated unit test code for display in the frontend chat UI

**How to Test**
docker compose up -d
then wait until docker logs testgen shows that the model has been fully pulled. Should print "success" in the console
If you call /generate-test before the model finishes loading, it may silently fail or return "Something went wrong"

**Future Improvments:** 
- Add timeout-based shutdowns via Kubernetes or other orchestration platform
- Improve speed of model call by experimenting with different local models
- Add more assignment examples to improve RAG quality
- Style AI output to be more readable (formatting, syntax highlighting, etc.)
- Lazy-load Ollama containers only when needed
- Add spinner/loading animation during AI response generation
- Add tests + response validation
- Further improve generate tests to be able to remember conversations

### DIFF
--- a/backend/routes/__init__.py
+++ b/backend/routes/__init__.py
@@ -13,3 +13,4 @@ def register_routes(app):
     app.register_blueprint(assignment)
     app.register_blueprint(regrade_request)
     app.register_blueprint(submission)
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,66 +1,74 @@
 version: "3.8"
 services:
-    # frontend:
-    #     build:
-    #         context: ./frontend/
-    #         target: prod
-    #     ports:
-    #         - "3000:3000"
-    #     volumes:
-    #         - ./frontend:/app
-    #         - /app/node_modules
-    #     stdin_open: true
-    #     command: npm start
-    # frontend-test:
-    #     build:
-    #         context: ./frontend/
-    #         target: test
-    #     volumes:
-    #         - /app/node_modules
-    #     environment:
-    #         - CI=true
-    #     command: npm run test
-    backend:
-        container_name: flask_container
-        build: 
-            context: "./backend/"
-            dockerfile: Dockerfile
-        environment:
-            FLASK_APP: app.py
-            FLASK_DEBUG: 1
-            PYTHONUNBUFFERED: 1
-        ports:
-            - "5001:5001"
-        extra_hosts:
-              - "host.docker.internal:host-gateway"
-        volumes:
-            - ./backend/:/usr/src/app
-            - /var/run/docker.sock:/var/run/docker.sock
+  # frontend:
+  #     build:
+  #         context: ./frontend/
+  #         target: prod
+  #     ports:
+  #         - "3000:3000"
+  #     volumes:
+  #         - ./frontend:/app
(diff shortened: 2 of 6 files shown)


shreyavarma99:
The model will probably take upwards of 2 mins to respond for now. We can work on speeding it up in the future.

yangshao05 (commented):
It runs well, successfully loaded the model and opened the dialog box. The issue on my end was that I needed to add `platform: linux/amd64` to the testgen container during the Docker Compose process to override the default architecture. This caused Docker to run the container under QEMU emulation, which slowed the model. As a result, I had to stop the process before receiving a response because it overloaded my CPU. If someone else could continue testing from here, that would be helpful.

I’m not sure if this override is needed for other Mac users as well. It’s likely the image was built only for amd64, while Mac uses arm64. If this issue occurs for others, perhaps we can consider publishing the image as multi-architecture so it can run natively on both amd64 and arm64.

shreyavarma99:
Thanks for pointing out that the test generation feature wasn’t working on your architecture. I’ve updated the microservice so it should now work on your setup. Could you pull the latest changes from the PR and test it on your Mac?
I don’t have a Mac to test with, so it’d be easier if you could try it out on your end. If it still doesn’t work, we may need to connect at some point to debug together.

yangshao05:
Thank you for the update. It’s working on my architecture now! Since the model was exceeding my total cpu, I had to stop it before getting a response. It might be helpful if someone else could test the output.

makyra:
I pulled testgen and tried running the system, but I get a network error. I am assuming that an additional URL has to be added for my frontend env to reach the testgen service on port 5002, but I am not sure if my understanding is correct. 
Am I testing this correctly? Thank you!

shreyavarma99:
I have a theory about why you might be getting that error. Remember at the start of the summer when we had to tweak your docker-compose.yml to use a different database name because of that weird issue? My current config doesn’t have those changes, so when you pulled, it may have overwritten yours.

You might want to redo that database config change and try connecting again. If it’s still acting up, we can hop on a quick call to debug together.

makyra:
For visibility, I had already messaged Shreya about this, but updating here as well:

After reviewing the yml files and diff-checking, I confirmed that the initial setup and database names are not the issue. The network error is occurring on my end because I accidentally broke my local database while trying to fix earlier migration bugs.

I am currently debugging this and will update once I resolve it and re-run the AI testgen. In the meantime, it would be greatly appreciated if someone else could confirm that this PR works successfully. Thank you!

PR #268, opened 2025-08-19
added linux docs 

WajeehJ:
pull request for linux docs (still differences between this and the linux branch so these docs will not necessarily work for main)


kiat:
Thanks!

PR #270, opened 2025-09-09
docs(readme): reorder backend .env before docker compose up

Nominose:
Reorders the README steps so that backend `.env` is set up
after frontend `.env` and before `docker compose up`.

Why:
Starting containers before defining `backend/.env` can cause confusing
errors (DB/JWT not loaded). This ordering makes the first run more reliable.

Changes:
 Moved backend `.env` section right after frontend `.env`

Scope: docs only; no code behavior change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,22 +37,7 @@ Note: For macOS make sure to turn off Airplay as it uses localport:5000 as well 
     REACT_APP_API_URL={where your backend is hosted}
     ```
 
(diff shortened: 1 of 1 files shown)


makyra (approved):
This sounds good to me! I am currently setting up codeAssist again, and I see why this could be a confusion point.

kiat:
Thanks!

PR #283, opened 2025-10-15
Add AI Settings

yangshao05:
Moved AI settings to separate page.

- Added a new instructor page AI Settings and moved the AI key UI there. It supports provider selection (OpenAI, Google Gemini, Anthropic), API key entry, temperature, and a “Test Connection” button.

Wired routing and sidebar:
- Added route: /aiSettings/:courseId
- Added sidebar entry with icon in ExpandedSidebar
- Removed the AI Integration card from Course Settings

Screenshot 2025-10-15 at 12 46 02

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -11,6 +11,7 @@ import RootSider from "./components/layout/sider";
 import "./mock";
 import InstructorDashboard from "./pages/instructor/dashboard";
 import CourseSettings from "./pages/instructor/courseSettings";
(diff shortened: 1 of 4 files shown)


kiat:
Can we separate the AI system from the model type. For example, we have OpenAI API with GPT4 or GPT5, or others. We can have a text field that we enter which specific model of the OpenAI we want to use.

kiat:
Thanks!

Issue #276, opened 2025-10-08
Updated User Role Doesn’t Persist or Reflect on Student’s End


yczhangcc:
Location: Instructor → Course Roster (Manage Enrollment)

Description: 

When an instructor changes a user’s role (e.g., Student → TA), the UI shows a toast “User role updated.”
  - The update is not saved permanently — after page refresh, navigation, or re-login, the role reverts to Student.
  
The search and filter controls on the roster page do not function correctly:
- Filtering by role or name does not update the table results.
- Resetting the filter does not reload the full enrollment list.

<img width="1820" height="866" alt="Image" src="https://github.com/user-attachments/assets/35c7059a-afd0-4d0a-8248-725bbf3965d7" />

PR #286, opened 2025-10-20
Fix #276 : updated user roles and enable search/reset filters on enrollment page 

yczhangcc:
Fixes Implemented
Updated /get_course_enrollment to join Enrollment and User tables.
Ensures that role updates (e.g., Student → TA) now persist after refresh or login.

Frontend:  
Fixed search and reset buttons in the Instructor 
Implemented working filters by role and username.
Table now displays correct, updated role information.

image

### DIFF
--- a/backend/routes/course.py
+++ b/backend/routes/course.py
@@ -420,25 +420,35 @@ def get_user_enrollments():
 @cross_origin()
 def get_course_enrollment():
     """
-    /get_course_enrollment gets all students enrolled in a course
-    Requires from the frontend a JSON containing:
-    @param course_id        the id of a course
+    Returns all users enrolled in a course,
+    showing their course-specific role from the Enrollment table.
     """
     course_id = request.args.get("course_id")
-    if not course_id or course_id == "":
+    if not course_id:
         raise BadRequestError("Missing course_id argument")
 
-    students = db.session.query(Enrollment.student_id).filter_by(course_id=course_id)
-    students = EnrollmentSchema().dump(students, many=True)
+    results = (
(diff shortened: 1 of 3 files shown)


EswarG8126 (commented):
[MY ONLY IMPORTANT COMMENT, (everything else was an ACCIDENT)] I have reviewed the code, ran the tests and came to one major conclusion. First off, the tests are not successful on my end. The major conclusion is that the errors appear to be in the sign up or log in modal files. There might be need for some changes. Will wait on others to comment since my system is acting buggy.

EswarG8126 (dismissed):
Look at my other comment, I would approve but there seems to be other issues on my end with the log in and sign up modals. This might just be on mine since I was working on another feature as well (messed up the flow of fetching).

yangshao05:
To filter by role, I need to reset each time. For example, if I filter by 'Instructor' first, then filter by 'Student', it will not list any student in the form, I have to reset and then filter by 'Student'. I think it's better to allow multiple filters without requiring reset.
Screenshot 2025-10-29 at 17 53 10
Screenshot 2025-10-29 at 17 53 19
Screenshot 2025-10-29 at 17 53 28

yczhangcc:
Thanks for the feedback! should have it fixed by the end of the week.

makyra (approved):
All works as expected, and I can see that the previous issues are resolved. I believe the next steps could be to extending capability of TAs so that they can grade or view student submissions.

krishaaapatel (approved):
Looks good on my side too!!

kiat:
Thanks!

PR #293, opened 2025-11-17
Fixed Failing Frontend Tests

makyra:
All 15 failing tests fixed, and now all 20 test suites and 45 tests passing. The following changes were made.
- Updated outdated tests:
  - Added missing mocks to prevent runtime crashes.
  - Updated tests to use axios, matching the current implementation, instead of the old global fetch.
  - Mocked API calls consistently for pages that rely on chained fetches.
 
- Fixed issues in code:
  - Cleaned up upload & submission flow.
  - Updated AssignmentModal to work with the axios-based service layer.
  - Ensured fetch calls and submission logic behave consistently with mocks.


yczhangcc (commented):
image

The frontend tests all pass now, but I just noticed that the backend tests are failing on my machine ,even on the main branch. I'm not sure if this is only happening on my end. If someone else could check the backend tests, that would be helpful.
image

makyra:
On my end, I do not have any backend issues at all, but let's see if other students have the issue as well. Thank you for the review!

kiat:
Please also review by running "make test" on your local clone.

krishaaapatel (commented):
On my side, frontend tests are passing, but backend tests aren't. Getting the same result as Cecilia.

**UPDATED- I think the PR looks good, as the PR fixed the issues for the frontend tests, and on my side all frontend tests passes now!

Screenshot 2025-11-20 at 5 53 59 PM
Screenshot 2025-11-20 at 5 54 12 PM

yangshao05 (approved):
Looks good to me.
It happens to me during the summer that backend tests always failed, to run the backend tests I need to change all ‘routes.user.db’ to ‘api.db’. Since our Github workflow runs the backend tests, and no backend changes made, it should be safe to merge.

kiat:
Thank you for your work!

Issue #291, opened 2025-11-07
Student Portal Displays Incorrect Assignment Release & Due Times


krishaaapatel:
When an assignment is created on instructor side at a specific time (ex: 12:40 PM CDT), the instructor view shows the correct timestamps. However, the student portal displays the due/release timestamps shifted by +6 hours (ex: 6:40 PM instead of 12:40 PM). This makes students see incorrect deadlines even though backend saved the correct time.

PR #294, opened 2025-11-19
Fix timestamp timezone handling (Issue #291)

krishaaapatel:
Fixed the timestamp issue where assignment due dates and release dates appeared correctly on the instructor side but showed a +6-hour shift on the student side.

The root cause was that the backend stored all assignment timestamps using `TIMESTAMP WITHOUT TIME ZONE`, which meant the database saved raw datetimes with no timezone information. As a result, the frontend interpreted these timestamps differently for instructors vs. students, causing the student dashboard to display incorrect local times.

The fix updates all assignment and submission-related datetime fields to use `TIMESTAMPTZ` (timestamp with time zone). This ensures the backend consistently stores all timestamps in UTC and returns them with proper timezone metadata. After migrating the database, both instructor and student portals now display the correct local times.

Also removed the temporary UTC hack on the instructor UI, since timezone conversion is now handled correctly by the backend. All timestamps are now stored in UTC and automatically converted to the user’s local timezone on both sides.

# NEED TO UPDATE THE LOCAL DB FOR THIS PR

need to change several datetime fields from TIMESTAMP WITHOUT TIME ZONE → TIMESTAMPTZ

docker exec -it postgres_container sh

psql -U postgres

\c codeassist;

ALTER TABLE assignments
  ALTER COLUMN due_date TYPE timestamptz USING due_date AT TIME ZONE 'UTC',
  ALTER COLUMN late_due_date TYPE timestamptz USING late_due_date AT TIME ZONE 'UTC',
  ALTER COLUMN published_date TYPE timestamptz USING published_date AT TIME ZONE 'UTC';

ALTER TABLE submissions
  ALTER COLUMN submitted_at TYPE timestamptz USING submitted_at AT TIME ZONE 'UTC';

ALTER TABLE assignment_extensions
  ALTER COLUMN release_date_extension TYPE timestamptz USING release_date_extension AT TIME ZONE 'UTC',
  ALTER COLUMN due_date_extension TYPE timestamptz USING due_date_extension AT TIME ZONE 'UTC',
  ALTER COLUMN late_due_date_extension TYPE timestamptz USING late_due_date_extension AT TIME ZONE 'UTC';


kiat:
Please review it before Thanksgiving!

yangshao05 (approved):
Looks good to me!

kiat:
Thank you!

PR #300, opened 2026-06-09
dynamically fetch models and use searchable dropdown 

Gaurav-021:
This change adds a hook that automatically calls OpenAI, Google, and Anthropic APIs to fetch the list of available models using your API key. Also replaces the plain text input field with a searchable component. 

Testing (video):
<video src="https://github.com/user-attachments/assets/7a26fc9a-b378-49fb-aef8-5e988e6e5f54" controls width="600"></video>


Gaurav-021:
Closing PR as Cecilla added this as part of #301

PR #301, opened 2026-06-10
Update multi-provider AI settings for course and assignment feedback

yczhangcc:
Summary
solve issue #310 #311 
This PR updates the AI feedback settings workflow for both course-level and assignment-level configuration.

Main changes:

- Added support for multiple AI providers:  OpenAI / ChatGPT,  Google Gemini,  Anthropic Claude
- Updated course-level AI settings to support provider-specific API keys and default model configuration.
- Added API key testing and selected model testing.
- Updated assignment-level AI feedback settings so assignments can either:
- Use course default AI settings, or Customize provider/model for that assignment only.
- Added default correctness-focused AI feedback prompt.
- Updated AI feedback generation to focus on correctness, debugging, failed tests, edge cases, and runtime behavior.
- Added model filtering logic for known unsupported or special-purpose models.
- Added documentation for how to use and test AI Settings.

Notes!!!!
The API key test and selected model test are intentionally separate.
A provider API key may be valid, but a selected model may still fail because some models are too advanced, deprecated, special-purpose, or require a different request format. Very normal API chaos, naturally.

Testing
OpenAI, Gemini, and Claude can all connect through the AI Settings page.
Users should click Fetch Models before choosing a model.
Users should click Test Selected Model before saving or using a model.
Some provider-returned models may not be usable for our AI feedback workflow.
If a model test fails, choose another available model and test again.

TEST NOTE:
docker compose down
docker compose up
Leave this terminal open so backend logs are visible.
If the database was already migrated before, you may not need to reset volumes. But if you see missing column/table errors, run migration again.
!!Run database migration 
Open a new terminal:
cd ~/Desktop/.../codeAssist
docker compose exec backend sh
flask db upgrade
python3 init_db.py
exit
Backend:
PYTHONPATH=. python -m pytest
Frontend:
npm test -- --watchAll=false

https://github.com/user-attachments/assets/803fc364-5883-4c22-a2c8-f3dae66ce287

### DIFF
--- a/backend/routes/course.py
+++ b/backend/routes/course.py
@@ -28,18 +28,35 @@ def is_supported_openai_model(model_id):
     """
     Filter OpenAI models shown in the dropdown.
 
-    Remove o-series models that fail with the current test request because
-    they do not support max_tokens in the chat completions call.
+    Keep standard text chat models that work with this app's chat completion
+    request shape. Remove o-series and special-purpose models that may appear
+    in the models API but fail with max_tokens, response_format, or text-only
+    chat completions.
     """
-    blocked_models = {
-        "o3-mini",
-        "o4-mini",
-    }
+    normalized_model_id = (model_id or "").lower()
+    blocked_prefixes = ("o1", "o3", "o4")
+    blocked_keywords = [
+        "audio",
+        "dall-e",
+        "embedding",
+        "image",
+        "instruct",
+        "moderation",
+        "preview",
+        "realtime",
+        "search",
+        "transcribe",
+        "tts",
+        "whisper",
+    ]
 
-    if model_id in blocked_models:
+    if normalized_model_id.startswith(blocked_prefixes):
         return False
 
-    return model_id.startswith("gpt-")
+    if any(keyword in normalized_model_id for keyword in blocked_keywords):
+        return False
+
+    return normalized_model_id.startswith("gpt-")
 
 
 def is_supported_gemini_model(model_id):
@@ -60,6 +77,9 @@ def is_supported_gemini_model(model_id):
         "learnlm",
         "deep-research",
         "antigravity",
+        "preview",
+        "exp",
+        "experimental",
     ]
 
     blocked_models = {
--- a/backend/routes/course.py
+++ b/backend/routes/course.py
@@ -1,6 +1,7 @@
 import uuid
 import os
 import csv
+import requests
 from flask import Blueprint, request, jsonify
 from flask_cors import cross_origin
 from werkzeug.utils import secure_filename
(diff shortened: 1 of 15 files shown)


frontend/src/pages/instructor/aiSettings/index.js
SashankN7, line 1: lgtm

SashankN7 (approved):
Looks good

DarioDeSilva (approved):
LGTM, removes the duplicated course-level model setting, keeps provider/credential behavior intact, and is low-risk.

frontend/src/pages/instructor/aiSettings/index.js
```diff
@@ -1,212 +1,489 @@
...
+                {!hasSavedKey && (
+                  <Alert
+                    type="warning"
+                    showIcon
+                    message={`${selectedProvider?.label} API key is not saved`}
+                    description="Add and save an API key before using this provider for AI feedback."
+                  />
+                )}
+              </Space>
             </Card>
```
Gaurav-021: Do we need an explicit "AI Setup Status" card? If the connection test succeeds and a default model is chosen, the setup status is implicitly complete. I know we previously directly queried provider APIs like Anthropic/OpenAI from the browser but I don't see a need to change up the ui.

frontend/src/pages/instructor/aiSettings/index.js
```diff
@@ -1,212 +1,489 @@
...
+                ]}
+              >
+                <Space.Compact style={{ width: "100%" }}>
+                  <Button
+                    onClick={handleFetchModels}
+                    loading={modelsLoading}
+                    style={{ width: 180 }}
+                  >
+                    Fetch Models
+                  </Button>
```
Gaurav-021, line 389: Instead of requiring the user to manually click the "Fetch Models" button, we could automate this fetch:

1. Fetch on load if a saved key is found in  loadCourseAiInfo()
2. Fetch automatically once the user finishes typing/pasting a new API key.

frontend/src/pages/instructor/aiSettings/index.js
```diff
@@ -1,212 +1,489 @@
...
+                  message={
+                    hasSavedKey
+                      ? `${selectedProvider?.label} API key is saved`
+                      : `${selectedProvider?.label} API key is not configured`
+                  }
+                  description={
+                    hasSavedKey
+                      ? "Saved key is loaded into the password field and hidden by default. Use the visibility icon to view it."
+                      : "Paste an API key below and save it for this provider."
+                  }
```
Gaurav-021, line 437: Sorry, this might just be being nitpicky but:

the helper text explaining how to view the hidden key is likely not needed (the eye logo should be enough).

backend/routes/course.py
```diff
@@ -511,6 +637,443 @@ def store_api_key():
...
+        db.session.commit()
+        return jsonify({"message": "AI settings updated successfully"}), 200
+    except Exception:
+        db.session.rollback()
+        raise InternalProcessingError("Failed to update AI settings")
+
+
+@course.route("/fetch_ai_models", methods=["POST"])
+@cross_origin()
+def fetch_ai_models():
```
Gaurav-021, line 693: It might be better to split this PR as it adds a lot of code complexity: keeping this PR focused purely on removing the duplicate UI and handling simple course-level API keys. 

Also, I see that you made a few big commits to this PR after you received approvals to merge; I am not too sure what the correct way to go about it is, but maybe you should contact your other reviewers and let them know to review these other changes?

Gaurav-021 (commented):
This is a massive upgrade over the previous approach!!

backend/routes/course.py
yczhangcc: That is fair. The PR got bigger after the professor meeting because the duplicate model setting was tied to the larger AI settings workflow, including course defaults, assignment overrides, provider keys, model fetching, and model testing.
I renamed the PR and already asked the other reviewers to review the latest changes again. Splitting it now would be difficult with my current branch state, so I plan to keep this together and move future polish to follow-up PRs.

frontend/src/pages/instructor/aiSettings/index.js
yczhangcc: agreed

frontend/src/pages/instructor/aiSettings/index.js
yczhangcc: Agreed. I removed the separate AI Setup Status card in the latest update and replaced it with smaller inline status tags for API key status and selected model test status. This keeps the UI closer to the original flow while still making debugging clearer when the API key exists but the selected model has not been tested yet.

frontend/src/pages/instructor/aiSettings/index.js
yczhangcc: Good point. I updated the flow so models refresh automatically when the page loads and a saved key exists. After saving a new key, the page reloads the course AI info, which also refreshes the model list.
I avoided fetching models while the user is typing the key because it could send incomplete keys and create noisy errors. If we want that later, an onBlur fetch would be safer than fetching on every change.

kiat:
Thanks everyone!

PR #305, opened 2026-06-16
Fix resubmit runtime issue (Fixes Issue #304)

### LINKED ISSUE #304: Bug: Resubmit Assignment button causes runtime error
### Description

When I click the **Resubmit Assignment** button on the assignment result page, the page crashes with an uncaught runtime error.

This looks like the upload component expects `fileList` to be an array, but it is receiving another type, possibly an object/string/null from the resubmit flow.

Image

khu-droid:
Upload and submission flows were unstable due to incorrect file state handling. This caused UI crashes and inconsistent behavior.

Frontend
- Fixed Ant Design Upload by using controlled `fileList`
- Ensured file state resets on:
  - modal close
  - file removal
  - successful upload
- Fixed submission history modal crash cases

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -496,8 +496,14 @@ def get_active_submission():
 
     if not assignment or not student:
       raise BadRequestError("not sufficient details")
+
+    try:
+        student_uuid = uuid.UUID(str(student))
+        assignment_uuid = uuid.UUID(str(assignment))
+    except (ValueError, TypeError):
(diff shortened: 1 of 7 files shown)


rosered123:
When I run a check, it seems to be failing.

rosered123 (approved):

DarioDeSilva (approved):

kiat:
Thank you everyone!

PR #306, opened 2026-06-16
Fix instructor dashboard metrics and assignment normalization

khu-droid:
Instructor dashboard displayed inconsistent or incorrect assignment metrics, including submission counts, graded percentages, and regrade totals. Sorting for assignments was also unreliable.

Backend
- Updated `/get_course_assignments` to return:
  - submissions (distinct students)
  - graded_count
  - total_students
  - regrades
  - corrected % graded calculation

Frontend
- Normalized assignment data across instructor dashboard and assignments pages
- Fixed sorting:
  - name sorting uses `localeCompare`
  - date sorting uses proper `Date` comparisons

Tests
- Updated tests for assignment metrics and normalization


kiat:
Please check the backend conflicts and fix them.

DarioDeSilva (changes requested):
Good work on normalizing the assignment data and fixing the sorters. There are a few issues to address before this can merge.

---

1. submission_count and graded_count missing active=True filter
   File: backend/routes/course.py, lines 620-632

   Both queries count all submissions, including old inactive ones. A
   student who resubmits will have their first scored (inactive) submission
   still counted as graded, even though their current active submission has
   no score. Every other submission query in submission.py uses active=True.
   The dashboard graded% will be higher than reality.

---

2. graded percentage can exceed 100
   File: backend/routes/course.py, line 643

   graded_count comes from the Submission table and includes students who
   may have since dropped the course. enrolled_student_count is the current
   live enrollment count. When graded_count > enrolled_student_count, the
   formula produces a number over 100. Ant Design's Progress bar will
   render as broken/overflowed for those assignments.

---

3. N+1 query problem — 3 DB queries per assignment inside a loop
   File: backend/routes/course.py, lines 617-644

   For each assignment the endpoint fires 3 separate COUNT queries
   (submissions, graded, regrades). A course with 30 assignments makes
   91 DB round-trips on every dashboard page load. All three can be
   replaced with two GROUP BY queries before the loop, reducing to
   3 total queries regardless of course size.

---

4. regrade_count includes regrade requests on inactive submissions
   File: backend/routes/course.py, line 633

   The join to Submission has no active=True filter. When a student
   resubmits after requesting a regrade, the original request stays
   attached to the now-inactive submission and still appears in the
   count. Instructors see phantom regrades for work that has already
   been superseded.

---

5. No null guard on RELEASED and DUE columns in the dashboard
   File: frontend/src/pages/instructor/dashboard/index.js, lines 36 and 42

   The render function calls moment(text).format(...).toUpperCase()
   without checking if text is null. When an assignment has no
   published_date or due_date, normalizeAssignmentForTable maps the
   field to null, and moment(null) in v2.x returns an invalid moment
   whose format() output is "INVALID DATE". Assignments.js on this
   same PR already has the null guard — the dashboard copy missed it.

---

6. initFetch has no try/catch
   File: frontend/src/pages/instructor/dashboard/index.js, line 78

   The PR removed the old fetchData wrapper which had a try/catch with
   console.error. The replacement async function has no error handling
   at any level and is called without .catch(). A network failure or
   5xx response silently leaves the dashboard empty with no message
   to the instructor.

---

7. graded percentage computed twice — backend and frontend
   File: frontend/src/utils/assignmentData.js, line 14

   The backend already sends a graded field (rounded percent).
   normalizeAssignmentForTable recomputes it from graded_count divided
   by total_students and overwrites the backend value unconditionally.
   When total_students is missing and falls back to submissions count,
   the denominator changes silently and the two sides disagree on the
   percentage shown.

---

8. getPasswordRules exported from the wrong file, no production callers
   File: frontend/src/utils/assignmentData.js, line 1

   This function has no relationship to assignment data. It is not
   imported by any production component anywhere in the codebase, only
   by its own test. When a developer needs password validation rules
   for a form they will not find it here and will write a duplicate
   elsewhere, causing the two implementations to drift.

---

9. Null dates sort to epoch instead of end of list
   File: frontend/src/pages/instructor/assignments/Assignments.js, line 35

   The sorter uses new Date(a.released || 0) — when released is null
   this becomes new Date(0), which is January 1 1970. Unpublished
   assignments sort before all real dates in ascending order rather
   than at the end, where most users would expect them.

---

10. Role string "student" hardcoded with no shared constant
    File: backend/routes/course.py, line 610

    The enrolled_student_count query filters by role="student" as a
    raw string literal. If this value is ever changed or normalised,
    the query silently returns 0 for every course and all assignments
    show 0% graded with no error raised. A shared constant used in
    one place would make the dependency explicit and catchable.

DarioDeSilva:
Also needs a rebase on main to pick up the better-mock import fix in frontend/src/mock/common.js. The file currently imports from mockjs which is no longer installed — main already has the one-line fix, the branch just needs to be updated to include it.

PR #309, opened 2026-06-17
docs: update pgadmin connection host to db in readme 

Gaurav-021:
On Linux, `host.docker.internal` does not resolve inside the pgAdmin container by default, causing a "connection is bad: name does not resolve" error in Step 7. This PR updates the host to `db`, which works cross-platform. 

I wasn't sure if an issue was necessary as it is a super small change.


yczhangcc (approved):

yczhangcc:
Thank you

PR #312, opened 2026-06-19
Fix/cascade delete user relationships

SashankN7:
## Problem
Deleting a user from the admin dashboard fails or leaves orphaned records because foreign keys lack ON DELETE CASCADE behavior.

### Backend
- **models.py**: Added ON DELETE CASCADE to all foreign keys referencing users.id, cascading through the full dependency chain (enrollments, submissions, code drafts, regrade requests, test case results, submission submitters, assignment extensions)
- **models.py**: `Course.instructor_id` uses ON DELETE RESTRICT to prevent accidental deletion of an instructor who owns active courses — deletion is blocked until courses are reassigned
- **user.py**: `delete_user` now catches `IntegrityError` from RESTRICT constraints and returns a proper 409 Conflict instead of a generic 500
- **errors.py**: Added `TooManyRequestsError` (429) for the code editor rate limiter
- **code_editor.py**: Session-based auth via `_verify_student` on all 6 code editor endpoints; enrollment verification on `ai_chat` and `run_code`; per-user sliding-window rate limiter (10 req/60s) on `run_code`; sanitized error responses
- **submission.py**: Session-based ownership verification (`_verify_student_owner`) on `get_submissions`, `upload_submission`, `get_latest_submission`, `get_active_submission`, `activate_submission`; authentication check on `rerun_submission_autograder`
- **__init__.py**: CORS locked to `FRONTEND_ORIGIN` with `supports_credentials=True`; `SECRET_KEY` required via env var
- New Alembic migration to update existing database constraints

### Frontend
- Enhanced deletion confirmation dialogs to explicitly warn about permanent data deletion (enrollments, submissions, grades, extensions, code drafts)
- Button text changed to "Yes, Delete Everything"

### Tests
- New integration tests (`test_cascade_delete_it.py`) with `PRAGMA foreign_keys=ON` to verify: cascade deletes remove enrollments/submissions when a student is deleted; deleting an instructor with active courses is blocked (RESTRICT); deleting one student does not affect other students
- Updated unit tests to mock session-based auth and exercise new error paths
- All unit and integration tests pass

### DIFF
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -1,4 +1,5 @@
 import os
+import logging
 from flask import Flask
 from flask_marshmallow import Marshmallow
 from flask_sqlalchemy import SQLAlchemy
@@ -31,8 +32,15 @@ def create_app(config_class='config.Config'):
     
     # Initialize the extensions with the app
     # Lock CORS to the actual frontend origin to prevent CSRF on credentialed requests.
-    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.
-    frontend_origin = os.environ.get('FRONTEND_ORIGIN', 'http://localhost:3000')
+    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.        frontend_origin = os.environ.get('FRONTEND_ORIGIN')
+    if not frontend_origin:
+        _logger = logging.getLogger(__name__)
+        _logger.warning(
+            "FRONTEND_ORIGIN environment variable is not set. "
+            "Defaulting to http://localhost:3000 for development only. "
+            "Set FRONTEND_ORIGIN to your actual frontend URL in production."
+        )
+        frontend_origin = 'http://localhost:3000'
     CORS(app, supports_credentials=True, origins=[frontend_origin])
     ma.init_app(app)
     db.init_app(app)
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -22,7 +22,7 @@ def create_app(config_class='config.Config'):
     app.config.from_object(config_class)
     
     # Initialize the extensions with the app
-    CORS(app)
+    CORS(app, supports_credentials=True)
     ma.init_app(app)
     db.init_app(app)
     migrate.init_app(app, db)
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -1,3 +1,5 @@
+import os
+import secrets
 from flask import Flask
 from flask_marshmallow import Marshmallow
 from flask_sqlalchemy import SQLAlchemy
@@ -13,16 +15,20 @@ migrate = Migrate()
 def create_app(config_class='config.Config'):
     # Create the Flask app instance
     app = Flask(__name__)
-    app.secret_key = 'codeassist'
 
-    # Load environment variables
+    # Load environment variables first so SECRET_KEY etc. are available
     load_dotenv()
+
+    # Use a random secret from env var; generate one if not set (dev only)
+    app.secret_key = os.environ.get('SECRET_KEY') or secrets.token_hex(32)
     
     # Load the configuration
     app.config.from_object(config_class)
     
     # Initialize the extensions with the app
-    CORS(app, supports_credentials=True)
+    # Lock CORS to the actual frontend origin(s) to prevent CSRF on credentialed requests
+    frontend_origin = os.environ.get('FRONTEND_ORIGIN', 'http://localhost:3000')
+    CORS(app, supports_credentials=True, origins=[frontend_origin])
     ma.init_app(app)
     db.init_app(app)
     migrate.init_app(app, db)
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -1,5 +1,4 @@
 import os
-import secrets
 from flask import Flask
 from flask_marshmallow import Marshmallow
 from flask_sqlalchemy import SQLAlchemy
@@ -18,15 +17,21 @@ def create_app(config_class='config.Config'):
 
     # Load environment variables first so SECRET_KEY etc. are available
     load_dotenv()
-
-    # Use a random secret from env var; generate one if not set (dev only)
-    app.secret_key = os.environ.get('SECRET_KEY') or secrets.token_hex(32)
     
-    # Load the configuration
+    # Load the configuration (provides defaults like SECRET_KEY for test config)
     app.config.from_object(config_class)
+
+    # SECRET_KEY must be set via env var or config class for session security.
+    app.secret_key = os.environ.get('SECRET_KEY') or app.config.get('SECRET_KEY')
+    if not app.secret_key:
+        raise RuntimeError(
+            "SECRET_KEY environment variable is required. "
+            "Set it to a random secret string (e.g. python -c 'import secrets; print(secrets.token_hex(32))')."
+        )
     
     # Initialize the extensions with the app
-    # Lock CORS to the actual frontend origin(s) to prevent CSRF on credentialed requests
+    # Lock CORS to the actual frontend origin to prevent CSRF on credentialed requests.
+    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.
     frontend_origin = os.environ.get('FRONTEND_ORIGIN', 'http://localhost:3000')
     CORS(app, supports_credentials=True, origins=[frontend_origin])
     ma.init_app(app)
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -32,7 +32,8 @@ def create_app(config_class='config.Config'):
     
     # Initialize the extensions with the app
     # Lock CORS to the actual frontend origin to prevent CSRF on credentialed requests.
-    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.        frontend_origin = os.environ.get('FRONTEND_ORIGIN')
+    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.
+    frontend_origin = os.environ.get('FRONTEND_ORIGIN')
     if not frontend_origin:
         _logger = logging.getLogger(__name__)
         _logger.warning(
--- a/backend/routes/code_editor.py
+++ b/backend/routes/code_editor.py
@@ -5,12 +5,14 @@ import subprocess
 import io
 import tarfile
 import threading
+import time
+import logging
 import requests
 from datetime import datetime, timezone
 from flask import Blueprint, request, jsonify, current_app
 from flask_cors import cross_origin
 from api import db
-from api.models import CodeDraft, Assignment, Submission, User, AssignmentExtension, Course
+from api.models import CodeDraft, Assignment, Submission, User, AssignmentExtension, Course, Enrollment
 from api.schemas import CodeDraftSchema, SubmissionSchema
 from util.errors import BadRequestError, NotFoundError, InternalProcessingError, SubmissionTimeoutError
 from util.url_utils import validate_ollama_url
@@ -23,8 +25,48 @@ from ai_feedback.integration import (
 )
 import docker
 
+logger = logging.getLogger(__name__)
+
 code_editor = Blueprint('code_editor', __name__)
 _docker_client = None
+
+# --- Rate limiting for run_code ---
+_run_code_timestamps = []
+_run_code_rate_lock = threading.Lock()
+_RUN_CODE_RATE_LIMIT = 10  # max requests
+_RUN_CODE_RATE_WINDOW = 60  # per 60 seconds
+
+
+def _check_run_code_rate_limit():
+    """Simple sliding-window rate limiter for run_code (thread-safe)."""
+    now = time.time()
+    with _run_code_rate_lock:
+        # Prune timestamps outside the window
+        while _run_code_timestamps and _run_code_timestamps[0] < now - _RUN_CODE_RATE_WINDOW:
+            _run_code_timestamps.pop(0)
+        if len(_run_code_timestamps) >= _RUN_CODE_RATE_LIMIT:
+            raise BadRequestError("Too many run requests. Please wait a moment and try again.")
+        _run_code_timestamps.append(now)
+
+
+def _verify_student(student_id):
+    """Verify that student_id refers to a valid user. Raises 400 if invalid."""
+    if not student_id:
+        raise BadRequestError("Missing student_id")
+    user = db.session.query(User).filter_by(id=student_id).first()
+    if not user:
+        raise BadRequestError("Invalid student_id: user not found")
+    return user
+
+
+def _verify_enrollment(student_id, course_id):
+    """Verify that a student is enrolled in the given course. Raises 403 if not."""
+    enrollment = db.session.query(Enrollment).filter_by(
+        student_id=student_id, course_id=course_id
+    ).first()
+    if not enrollment:
+        raise BadRequestError("You are not enrolled in this course")
+    return enrollment
 
 def get_docker_client():
     global _docker_client
@@ -47,8 +89,10 @@ def save_code_draft():
     file_name = data.get("file_name", "solution.py")
     auto_saved = data.get("auto_saved", False)
 
-    if not student_id or not assignment_id or content is None:
-        raise BadRequestError("Missing required fields: student_id, assignment_id, content")
+    _verify_student(student_id)
+
+    if not assignment_id or content is None:
+        raise BadRequestError("Missing required fields: assignment_id, content")
 
     if len(content) > 100000:
         raise BadRequestError("Code content exceeds maximum length of 100KB")
@@ -92,8 +136,10 @@ def get_code_drafts():
     assignment_id = request.args.get("assignment_id")
     condensed = request.args.get("condensed", "false").lower() == "true"
 
-    if not student_id or not assignment_id:
-        raise BadRequestError("Missing student_id or assignment_id")
+    _verify_student(student_id)
+
+    if not assignment_id:
+        raise BadRequestError("Missing assignment_id")
 
     drafts = (
         db.session.query(CodeDraft)
@@ -128,8 +174,10 @@ def get_latest_draft():
     student_id = request.args.get("student_id")
     assignment_id = request.args.get("assignment_id")
 
-    if not student_id or not assignment_id:
-        raise BadRequestError("Missing student_id or assignment_id")
+    _verify_student(student_id)
+
+    if not assignment_id:
+        raise BadRequestError("Missing assignment_id")
 
     draft = (
         db.session.query(CodeDraft)
@@ -158,8 +206,10 @@ def submit_code():
     content = data.get("content")
     file_name = data.get("file_name", "solution.py")
 
-    if not student_id or not assignment_id or content is None:
-        raise BadRequestError("Missing required fields: student_id, assignment_id, content")
+    _verify_student(student_id)
+
+    if not assignment_id or content is None:
+        raise BadRequestError("Missing required fields: assignment_id, content")
 
     if len(content) > 100000:
         raise BadRequestError("Code content exceeds maximum length of 100KB")
@@ -175,6 +225,10 @@ def submit_code():
 
     if not assignment.published:
         raise BadRequestError("Assignment is not published yet.")
+
+    # Enforce enable_code_editor server-side
+    if not assignment.enable_code_editor:
+        raise BadRequestError("Code editor is not enabled for this assignment.")
 
     # Check due dates (same logic as upload_submission)
     extension = db.session.query(AssignmentExtension).filter_by(
@@ -272,6 +326,13 @@ def submit_code():
             timeout=assignment.autograder_timeout,
         )
     except subprocess.TimeoutExpired:
+        # Ensure container cleanup on timeout
+        try:
+            container.stop()
+            container.remove()
+        except Exception:
+            pass
+
         timeout_result = {
             "tests": [{
                 "name": "Submission Timeout",
@@ -669,8 +730,11 @@ def run_code():
     content = data.get("content")
     file_name = data.get("file_name", "solution.py")
 
-    if not student_id or not assignment_id or content is None:
-        raise BadRequestError("Missing required fields: student_id, assignment_id, content")
+    _verify_student(student_id)
+    _check_run_code_rate_limit()
+
+    if not assignment_id or content is None:
+        raise BadRequestError("Missing required fields: assignment_id, content")
 
     if len(content) > 100000:
         raise BadRequestError("Code content exceeds maximum length of 100KB")
@@ -726,9 +790,10 @@ def run_code():
             run_image, content, file_name, timeout,
         )
     except Exception as e:
-        print(f"RUN_CODE _run_code_in_container error: {e}", flush=True)
+        # Log details server-side but return sanitized error to client
+        logger.error(f"RUN_CODE error: {e}", exc_info=True)
         return jsonify({
-            "output": f"Error: {type(e).__name__}: {str(e)}",
+            "output": "An error occurred while running your code. Please try again.",
             "programOutput": "",
             "passed": False,
             "score": 0,
@@ -797,8 +862,10 @@ def ai_chat():
     user_message = data.get("message")
     code = data.get("code", "")
 
-    if not student_id or not user_message:
-        raise BadRequestError("Missing student_id or message")
+    _verify_student(student_id)
+
+    if not user_message:
+        raise BadRequestError("Missing message")
 
     if not assignment_id:
         raise BadRequestError("Missing assignment_id")
@@ -809,6 +876,9 @@ def ai_chat():
         raise NotFoundError("Assignment not found")
     if not assignment.ai_feedback_enabled:
         raise BadRequestError("AI chat is not enabled for this assignment.")
+
+    # Verify student is enrolled in the course to prevent quota abuse
+    _verify_enrollment(student_id, assignment.course_id)
 
     course = db.session.query(Course).filter_by(id=assignment.course_id).first()
     if not course:
@@ -832,7 +902,7 @@ def ai_chat():
         reply = _get_ai_chat_reply(provider, api_key, client, user_prompt, model, temperature)
     except Exception as e:
         error_msg = str(e)
-        print(f"AI_CHAT error: {e}", flush=True)
+        logger.error(f"AI_CHAT error: {e}", exc_info=True)
         if "insufficient_quota" in error_msg or "429" in error_msg:
             raise BadRequestError("AI service quota exceeded. Please contact your instructor to update the API key.")
         elif "invalid_api_key" in error_msg or "401" in error_msg:
--- a/backend/routes/code_editor.py
+++ b/backend/routes/code_editor.py
@@ -41,10 +41,16 @@ def _check_run_code_rate_limit(student_id):
     """Per-user sliding-window rate limiter for run_code (thread-safe)."""
     now = time.time()
     with _run_code_rate_lock:
-        timestamps = _run_code_timestamps.setdefault(student_id, [])
+        timestamps = _run_code_timestamps.get(student_id, [])
         # Prune timestamps outside the window
         while timestamps and timestamps[0] < now - _RUN_CODE_RATE_WINDOW:
             timestamps.pop(0)
+        # Evict idle keys to prevent memory leak
+        if not timestamps:
+            _run_code_timestamps.pop(student_id, None)
+            # Record this request as the first in the window
+            _run_code_timestamps[student_id] = [now]
+            return
         if len(timestamps) >= _RUN_CODE_RATE_LIMIT:
             raise BadRequestError("Too many run requests. Please wait a moment and try again.")
         timestamps.append(now)
@@ -75,7 +81,7 @@ def _verify_enrollment(student_id, course_id):
         student_id=student_id, course_id=course_id
     ).first()
     if not enrollment:
-        raise BadRequestError("You are not enrolled in this course")
+        raise ForbiddenError("You are not enrolled in this course")
     return enrollment
 
 def get_docker_client():
@@ -239,6 +245,9 @@ def submit_code():
     # Enforce enable_code_editor server-side
     if not assignment.enable_code_editor:
         raise BadRequestError("Code editor is not enabled for this assignment.")
+
+    # Verify student is enrolled in the course
+    _verify_enrollment(student_id, assignment.course_id)
 
     # Check due dates (same logic as upload_submission)
     extension = db.session.query(AssignmentExtension).filter_by(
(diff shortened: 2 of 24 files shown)


backend/routes/code_editor.py
```diff
@@ -0,0 +1,620 @@
...
+
+@code_editor.route('/save_code_draft', methods=["POST"])
+@cross_origin()
+def save_code_draft():
+    """
+    Save a code draft (auto-save or manual save).
+    Expects JSON: { student_id, assignment_id, content, file_name?, auto_saved? }
+    """
+    data = request.json
+    student_id = data.get("student_id")
```
DarioDeSilva, line 100: **[SECURITY] Missing authentication on all code editor endpoints** — any caller can act as any student

All six endpoints in  (, , , , , `ai_chat`) accept `student_id` directly from the request body or query string with no session, JWT, or `@login_required` check. Compare to `upload_submission` in `submission.py` which has the same pattern but at minimum inherits any gateway-level auth.

**Failure scenario:** An unauthenticated attacker POSTs to `/submit_code` with `student_id=<victim>` and `assignment_id=<any published assignment>`. A graded submission is recorded under the victim's account — no credentials required. Similarly, any caller can drain another student's draft history via `GET /get_code_drafts?student_id=<anyone>`.

backend/routes/code_editor.py
```diff
@@ -0,0 +1,620 @@
...
+            file_name=file_name,
+            student_id=student_id,
+            assignment_id=assignment_id,
+            student_code_file=content.encode('utf-8'),
+            results=None,
+            score=None,
+            execution_time=0.0,
+            submitted_at=datetime.now(timezone.utc),
+            active=True,
+            completed=True,
```
DarioDeSilva, line 305: **[SECURITY] run_code spawns Docker containers with no rate limiting — unauthenticated DoS**

`run_code` calls `get_docker_client().containers.run(...)` inside a request handler that has no auth check and no rate limit. Any IP can POST to `/run_code` and spawn a Docker container.

**Failure scenario:** An attacker sends 500 concurrent POST requests to `/run_code` with arbitrary `student_id` values. Each request spawns a separate Docker container running the autograder image. The host exhausts CPU, memory, and Docker socket capacity, taking down the entire grading service until containers are manually cleaned up.

backend/routes/code_editor.py
```diff
@@ -0,0 +1,620 @@
...
+            return jsonify({
+                "output": result_data.get("output", ""),
+                "passed": result_data.get("score", 0) > 0,
+                "score": result_data.get("score", 0),
+                "tests": result_data.get("tests", []),
+            }), 200
+        else:
+            return jsonify({
+                "output": "Could not read results",
+                "passed": False,
```
DarioDeSilva: **[SECURITY] Internal exception details leaked to client via run_code (HTTP 200)**

The `except Exception as e` block returns `f"Error: {type(e).__name__}: {str(e)}"` in the response body with HTTP 200. This exposes internal infrastructure details.

**Failure scenario:** A Docker API error (e.g., image not found, socket permission denied) produces a response like `{"output": "Error: DockerException: Error while fetching server API version: HTTPConnectionPool(host=..., port=...)", "passed": false}`. This reveals Docker socket configuration, internal hostnames, and image names to any caller who can trigger an error. The HTTP 200 also means error-monitoring tooling keyed on status codes will not alert on these failures.

backend/routes/code_editor.py
```diff
@@ -0,0 +1,620 @@
...
+    if len(content) > 100000:
+        raise BadRequestError("Code content exceeds maximum length of 100KB")
+
+    # Check assignment exists and is published
+    assignment = db.session.query(Assignment).filter_by(id=assignment_id).first()
+    if not assignment:
+        raise NotFoundError("Assignment not found")
+
+    if not assignment.published:
+        raise BadRequestError("Assignment is not published yet.")
```
DarioDeSilva, line 238: **[BUG] submit_code does not check assignment.enable_code_editor**

`submit_code` checks `assignment.published` (line 182) but never checks `assignment.enable_code_editor`. The `enable_code_editor` column was added in migration `b2c3d4e5f6a7` precisely to let instructors gate which assignments allow browser-based submission. That gate is enforced only in the frontend UI, not in the API.

**Failure scenario:** An instructor creates an assignment with `enable_code_editor=False` (intending students to upload a structured zip). A student who knows the API POSTs directly to `/submit_code` and has their code graded — bypassing the file-upload requirement and any validation the autograder expects from a zip structure.

Suggested fix: add after line 220:
```python
if not assignment.enable_code_editor:
    raise BadRequestError("Code editor is not enabled for this assignment.")
```

backend/routes/code_editor.py
```diff
@@ -0,0 +1,620 @@
...
+
+
+@code_editor.route('/ai_chat', methods=["POST"])
+@cross_origin()
+def ai_chat():
+    """
+    Chat with AI about the current code.
+    Expects JSON: { student_id, assignment_id, message, code }
+    Returns: { reply: "..." }
+    """
```
DarioDeSilva, line 877: **[SECURITY] ai_chat burns instructor OpenAI quota with no enrollment verification**

`ai_chat` verifies `assignment.ai_feedback_enabled` and that the course has an OpenAI API key, but never checks that `student_id` is enrolled in the course. Combined with the missing auth (see auth finding), any unauthenticated party who can enumerate `assignment_id` values can trigger API calls against the instructor's key.

**Failure scenario:** An attacker scripts 10,000 POST requests to `/ai_chat` with a valid `assignment_id` and arbitrary `student_id` values. Each call fires an OpenAI API request at `max_tokens=500`. The instructor's API key hits its rate limit or runs out of quota, and they receive a large unexpected bill with no log of who made the calls.

frontend/src/components/VersionHistoryModal.js
```diff
@@ -0,0 +1,137 @@
...
+import { Modal, List, Tag, Button, Spin, Empty, message } from "antd";
+import { HistoryOutlined, FileOutlined, ClockCircleOutlined } from "@ant-design/icons";
+import moment from "moment";
+
+/**
+ * VersionHistoryModal — shows a list of auto-saved and manual draft versions.
+ *
+ * Props:
+ *   open             – boolean
+ *   onCancel         – () => void
```
DarioDeSilva, line 12: **[SECURITY] Bare fetch() bypasses auth layer — students can read each other's code drafts**

`VersionHistoryModal` uses `fetch()` directly instead of the `service()` axios helper that the rest of the frontend uses. Since the backend `get_code_drafts` endpoint has no auth check and accepts `student_id` as a plain query param, this creates a horizontal privilege escalation vulnerability.

**Failure scenario:** Student A opens DevTools and runs:
```js
fetch(`${API_URL}/get_code_drafts?student_id=<Student_B_id>&assignment_id=<id>&condensed=true`)
  .then(r => r.json()).then(console.log)
```
The backend returns Student B's full draft history — every auto-save and manual save — exposing their code before the submission deadline.

Fix: Use `service("get_code_drafts", { student_id, assignment_id, condensed: "true" })` here, and add server-side session verification that the requesting user's session matches the requested `student_id`.

DarioDeSilva (changes requested):
Requesting changes because this PR introduces several correctness and security regressions that should be fixed before merge.

**Missing server-side auth on code editor endpoints** (`backend/routes/code_editor.py`)
`student_id` is accepted from the request body without verifying it against the authenticated session/user. This allows a caller to potentially submit as another student, access or modify another student’s draft/submission data, and consume the project/instructor OpenAI quota. This is especially risky for `/run_code`, which can spawn Docker containers and does not appear to have rate limiting.

**Container leak on timeout** (`backend/routes/code_editor.py:699`)
When the autograder times out, `SubmissionTimeoutError` is raised before the container cleanup path runs. The Docker container may never be stopped or removed, which can leak resources over time.

**Late submission window is broken** (`frontend/src/pages/assignments/index.js:159`)
`lateDueDateTime` was removed without an equivalent replacement. As a result, students who are inside a valid late-submission window can still see “Due Date Has Passed” immediately after the regular deadline.

**Wrong state updated for late extension** (`frontend/src/pages/codeEditor/index.js:90`)
`setDueDate(extRes.data.late_due_date_extension)` should likely be `setLateDueDate(...)`. Updating the regular due date with the late extension corrupts the deadline logic for students with late-deadline extensions.

**`moment(null)` produces invalid date behavior** (`frontend/src/pages/codeEditor/index.js:85`)
If `late_due_date` is `null`, `lateDueDate` becomes invalid/falsy. This can block late submissions even when `late_submission=true`.

**`enable_code_editor` is not enforced server-side** (`backend/routes/code_editor.py:555`)
The flag appears to only be checked in the frontend. Students can still POST directly to `/submit_code` even if the instructor disabled the code editor. This needs to be enforced in the backend as well.

**`undefined.py` filename bug** (`frontend/src/pages/codeEditor/index.js:82`)
When the assignment name is missing, `undefined + ".py"` produces `"undefined.py"`, which is truthy, so the `|| "solution.py"` fallback never runs.

**Internal errors exposed to client** (`backend/routes/code_editor.py:924`)
The catch-all exception handler returns `type(e).__name__` and `str(e)` to the client. This can leak internal details such as Docker paths, container names, or backend implementation details. The API should return a generic error message to students and log the detailed error server-side.

SashankN7:
**Security Fixes**
> - Auth checks on all 6 code editor endpoints ( _verify_student ) — no more accepting any  student_id  without validation
> - Enrollment verification on  ai_chat  — prevents unenrolled users from draining instructor OpenAI quota
> - Rate limiting on  run_code  — 10 requests per 60 seconds, thread-safe, prevents Docker container DoS
> - Error sanitization on  run_code  — internal details (Docker paths, hostnames, image names) logged server-side only; client gets a generic message
> -  enable_code_editor  enforced server-side —  submit_code  now rejects submissions if the code editor isn't enabled for the assignment
> - VersionHistoryModal — replaced raw  fetch()  with authenticated  service()  helper so students can't read each other's drafts


 **Bug Fixes**
> - Container leak on timeout —  submit_code  now stops and removes the Docker container before raising  SubmissionTimeoutError 
> - Late extension state — fixed  setDueDate  →  setLateDueDate  so late extensions actually apply
> -  moment(null)  guard — prevents invalid date behavior that blocked late submissions
> -  undefined.py  filename — falls back to  "solution.py"  when assignment name is missing
> - Late submission window — restored the check so students in a valid late window aren't incorrectly blocked


 **Tests**
> - Updated  test_code_editor.py  to mock  _verify_student / _verify_enrollment  in non-auth tests
> - All 154 backend unit tests pass

SashankN7:
Fixed the merge conflict as well

DarioDeSilva:
Good job addressing all the previous feedback.

Two things still need fixing before this can be approved:

**1. `_verify_student` is not real authentication — horizontal privilege escalation still possible**

`code_editor.py` — `_verify_student()`

```python
def _verify_student(student_id):
    user = db.session.query(User).filter_by(id=student_id).first()
    if not user:
        raise BadRequestError("Invalid student_id: user not found")
    return user
```

This only confirms a user with that ID exists — it does not verify the authenticated session belongs to that user. Any caller can still pass `student_id=<victim>` and submit, read drafts, or trigger grading on their behalf. The fix needs to compare the authenticated session identity (e.g. `current_user.id`) against the `student_id` in the request. Without that, all six endpoints remain open to horizontal privilege escalation.

**2. Rate limiter is a global bucket — one student can block everyone**

`code_editor.py` — `_check_run_code_rate_limit()`

```python
_run_code_timestamps = []   # shared across ALL requests
_RUN_CODE_RATE_LIMIT = 10   # 10 total, not 10 per user
```

The 10 req/60s limit is shared globally. A single student running their code 10 times in a minute exhausts the entire budget, blocking every other student for the rest of that window. The timestamps list needs to be keyed by `student_id` (or session/IP) — a dict of per-user timestamp lists is the standard fix.

SashankN7:
**Problem 1: "Anyone can pretend to be anyone"**

**Before:** The code editor endpoints (save, submit, run, etc.) received a  student_id  in the request body and just checked "does a user with this ID exist in the database?" If yes, it did whatever that  student_id  asked for. So if I sent a request saying  student_id = <your ID> , the server would just accept it — even though I'm not you.

**After:** When you log in, the server now saves your user ID in a session cookie (like a wristband at a concert). Every time you hit a code editor endpoint, the server checks: "does the wristband match the  student_id  you're claiming?" If they don't match, it says "Nope, you can only access your own stuff" and blocks the request.
In short: Before, anyone could pass anyone's ID. Now, the server remembers who you are from login and verifies it matches.

**Problem 2: "One student can block everyone from running code"**

**Before:** There was a shared bucket of 10 "Run" requests per minute for all students combined. If Student A runs their code 10 times in a minute, every other student is blocked for the rest of that minute.

**After:** Each student now has their own bucket of 10 runs per minute. Student A's 10 runs only affect Student A. Student B still has their own 10 runs available.
In short: Before, one person could hog all the runs. Now, everyone gets their own fair share.

What was changed (5 files):
1.  user.py  (login) — When you log in (email/password or Google), the server now remembers your user ID in a session cookie
2.  code_editor.py  — The "verify student" check now compares your session cookie against the  student_id  you're claiming. Rate limiter switched from one shared list to a per-student dictionary
3.  __init__.py  — Turned on CORS credentials so the browser sends the session cookie with requests
4.  frontend/services/index.js  — Told axios (the HTTP client) to send cookies with every request
5.  test_code_editor.py  — Updated tests to set up a session before making requests, since auth is now enforced

DarioDeSilva:
Thanks for addressing the two issues from the last review — the `setLateDueDate` fix in `codeEditor/index.js` and the `inLateWindow` branch in `assignments/index.js` are both correctly resolved.

That said, it looks like the new session-based auth layer introduced to support this fix has brought along a few problems of its own that need to be sorted before this can merge.

**Security concerns (blocking):**

- **CORS misconfiguration** (`api/__init__.py`): Changing to `CORS(app, supports_credentials=True)` without also specifying `origins=` causes Flask-CORS to reflect any incoming `Origin` header back as `Access-Control-Allow-Origin`. In practice this means every origin is approved for credentialed requests, which opens up CSRF on all the newly protected endpoints. Need to lock this down to the actual frontend origin.

- **Hardcoded `secret_key`** (`api/__init__.py`): `secret_key = 'codeassist'` was harmless before, but now that the session is the sole auth gate on `submit_code`, `run_code`, `save_code_draft`, etc., it's a real problem. Anyone with repo access can forge a valid session cookie for any student using `flask-unsign`. This needs to be a randomly generated secret injected via environment variable.

- **Instructor cascade** (`models.py`): `ondelete="CASCADE"` on `courses.instructor_id` is probably not the intended behaviour. Deleting an instructor account will silently cascade through courses → assignments → submissions, wiping every student's grades and submission history for all courses they taught. CASCADE makes sense for student-owned rows, but for `instructor_id` the safer choice is `RESTRICT` (block deletion until courses are reassigned) or `SET NULL` with `nullable=True` (preserve the course record, just clear the FK).

**Correctness issues:**

- **Missing enrollment check in `run_code` and `submit_code`**: `ai_chat` enforces enrollment, but the two higher-impact operations don't. An authenticated student who knows an assignment ID from a course they're not enrolled in can spin up Docker runs and create graded submission records.

- **`enable_code_editor` not checked in `run_code`**: `submit_code` correctly rejects requests when the flag is off, but `run_code` has no equivalent guard, so disabling the code editor only half-works.

- **Wrong error class in `_verify_enrollment`**: The docstring says "Raises 403" but the code raises `BadRequestError` (400). `ForbiddenError` is already imported — just needs a swap.

**Minor issues:**

- The rate limiter in `_check_run_code_rate_limit` is per-process, so with multiple Gunicorn workers the effective per-student limit is `10 × worker_count`. Needs a shared store (Redis or similar) to actually hold.
- The same rate limiter never evicts idle keys — after a student's timestamps age out, the empty list stays in the dict forever. A simple `if not timestamps: del _run_code_timestamps[student_id]` after the prune loop fixes it.

DarioDeSilva (changes requested):
Good work on the core cascade delete changes and the date null-check bug fixes. There are a few issues that need addressing before this can merge.

---

### `@cross_origin()` defeats the new session auth entirely

**`backend/routes/code_editor.py` lines 89, 137, 178, 206, 612, 750** (and every other route file)

The PR adds `withCredentials: true` to the frontend axios instance and configures `CORS(app, supports_credentials=True, origins=[frontend_origin])` globally. However, every route in the project uses `@cross_origin()` with no arguments. Flask-CORS's global after_request hook checks whether `Access-Control-Allow-Origin` is already set and skips itself if it is. The per-route `@cross_origin()` runs first and sets `Access-Control-Allow-Origin: *` with no credentials header. The global config never fires.

Concrete effect:
- Login routes (`/user_login`, `/google_login`) return `Allow-Origin: *`, so the browser blocks the `Set-Cookie` response and the session cookie is never stored. No user can authenticate.
- All six code_editor routes fail CORS on every credentialed request.

Fix: Either remove all `@cross_origin()` decorators and let the global `CORS(app, ...)` handle everything, or pass the right options to each decorator: `@cross_origin(origins=[frontend_origin], supports_credentials=True)`.

---

### Migration applies `CASCADE` to `courses.instructor_id` but model says `RESTRICT`

**`backend/migrations/versions/c4d5e6f7a8b9_...py` line 33** vs **`backend/api/models.py` line 22**

The migration runs:
```python
op.create_foreign_key('courses_instructor_id_fkey', 'courses', 'users', ['instructor_id'], ['id'], ondelete='CASCADE')
```
But the model declares:
```python
instructor_id = db.Column(..., db.ForeignKey("users.id", ondelete="RESTRICT"), ...)
```
After the migration runs, deleting an instructor silently cascades and deletes all their courses, assignments, submissions, and student grades with no error. The admin confirmation dialog does not warn about this. Pick one and make both match.

---

### Fallback `secret_key` regenerates on every restart

**`backend/api/__init__.py` line 22**

```python
app.secret_key = os.environ.get('SECRET_KEY') or secrets.token_hex(32)
```
When `SECRET_KEY` is not set, every process start generates a new key. In a multi-worker deployment (gunicorn -w 4), each worker gets a different key and roughly 3 out of 4 authenticated requests return 403. On every restart, all logged-in users are silently logged out. `SECRET_KEY` should be a required env var rather than an optional one with a random fallback.

---

### `_verify_student` queries the DB before checking the session

**`backend/routes/code_editor.py` line 336**

The function queries the User table first, then checks the session. An unauthenticated caller gets `400 "Invalid student_id: user not found"` for an unknown UUID, or `403 "Not authenticated"` for a real user, leaking which UUIDs are valid. Check the session first (in-memory, free), then query the DB only if authenticated.

---

### Rate limiter returns HTTP 400 instead of 429

**`backend/routes/code_editor.py` line 324**

`BadRequestError` maps to 400 (bad input). Rate limiting should return 429 Too Many Requests. Adding a `TooManyRequestsError(status_code=429)` to `util/errors.py` would fix this with one line.

---

### `FRONTEND_ORIGIN` not set in production defaults to `localhost`

**`backend/api/__init__.py` line 30**

```python
frontend_origin = os.environ.get('FRONTEND_ORIGIN', 'http://localhost:3000')
```
Without this env var in production, the server only allows requests from localhost. Consider making it a required env var or documenting it clearly in the README.

DarioDeSilva:
Also needs a rebase on main to pick up the better-mock import fix in frontend/src/mock/common.js. The file currently imports from mockjs which is no longer installed — main already has the one-line fix, the branch just needs to be updated to include it.

frontend/src/pages/admin/instructors/manage.js
```diff
@@ -112,10 +112,10 @@ export default function AdminInstructorManage() {
           <Button onClick={() => navigate("/admin/instructors")}>Cancel</Button>
           <Popconfirm
             title="Remove Instructor"
-            description="This will completely remove the instructor from the system. All data including courses will be deleted. Are you sure?"
+            description="This will permanently delete the instructor and ALL associated data, including courses they teach, assignments, student submissions, and grades. This action cannot be undone. Are you sure?"
```
DarioDeSilva: The dialog now says "ALL associated data, including courses they teach, assignments, student submissions, and grades" but `courses.instructor_id` is `RESTRICT` in both the model and the migration. If the instructor has any courses, the DELETE will fail with an IntegrityError and return a generic 500 -- the admin clicks "Yes, Delete Everything" and gets "Failed to remove instructor." Either flip the FK to `CASCADE` so the dialog is accurate, or keep `RESTRICT` and update the dialog to tell the admin they need to remove the instructor's courses first. Either is fine, they just need to match.

backend/api/__init__.py
```diff
@@ -13,16 +14,26 @@
...
+        raise RuntimeError(
+            "SECRET_KEY environment variable is required. "
+            "Set it to a random secret string (e.g. python -c 'import secrets; print(secrets.token_hex(32))')."
+        )
     
     # Initialize the extensions with the app
-    CORS(app)
+    # Lock CORS to the actual frontend origin to prevent CSRF on credentialed requests.
+    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.
+    frontend_origin = os.environ.get('FRONTEND_ORIGIN', 'http://localhost:3000')
```
DarioDeSilva: Carry-over from round 2: if `FRONTEND_ORIGIN` is not set in production, this defaults to `localhost:3000` and CORS blocks every frontend request silently. Same fix as `SECRET_KEY` would work -- raise a `RuntimeError` if the env var is missing, or at least log a warning at startup.

DarioDeSilva (changes requested):
Good progress on round 3 -- the RESTRICT/CASCADE alignment, secret key hardening, session-before-DB ordering, 429 rate limit status, and all the `@cross_origin()` removals are all correctly addressed. The mock import rebase is in too. Two things left before this can merge.

DarioDeSilva:
This PR currently has merge conflicts with main — GitHub is showing it as CONFLICTING. Could you please rebase on main and resolve the conflicts? Once the diff is clean I'll do another review pass on the two remaining items.

DarioDeSilva:
Thanks for iterating through all the review rounds, this is close. Two things I'd like resolved before merge, plus one open question:

**Blocking:**
1. **Cascade behavior is currently unverified by tests.** SQLite (used by `TestConfig`) doesn't enforce foreign keys without a `PRAGMA foreign_keys=ON`, which isn't set anywhere, so the unit suite can't be exercising real cascade deletes, and `test_delete_user_integration` only deletes a bare user with no related rows. Could you add an integration test that creates a course/enrollment/submission chain, deletes the user, and asserts the related rows are actually gone, plus a test confirming that deleting an instructor with active courses is rejected (RESTRICT)? Right now the PR's core claim isn't covered by CI.
2. **PR description doesn't match the code.** It says cascading goes "through the full dependency chain (courses → assignments → submissions)," but `Course.instructor_id` is `RESTRICT`, not `CASCADE`: deleting an instructor with courses is blocked, not cascaded. Can you update the description so it doesn't mislead future readers?

**Worth a quick answer, not necessarily a blocker:**
3. `delete_user` catches the RESTRICT `IntegrityError` as a generic 500 ("Error deleting user"). The frontend disables the delete button client-side when an instructor has courses, but the API itself doesn't give a clear message if hit directly. Want to catch that case and return a proper 4xx, or is the frontend guard considered sufficient for now?
4. `_verify_student` (session-based ownership check) got added to all of `code_editor.py`, but `submission.py`'s `get_submissions`/`upload_submission`/`get_latest_submission` still take `student_id` straight from the request with no session check. Same class of issue: is that tracked separately, or should it ride along here?

Nice cleanup on the CORS lockdown and mandatory `SECRET_KEY`, good improvements. Once 1 and 2 are addressed I think this is good to merge.

DarioDeSilva:
Good progress on the previous rounds. Confirming what's resolved and flagging one new blocking issue found while re-reviewing.

**Resolved from the last round:**

1. Cascade behavior is now covered by `test_cascade_delete_it.py`, with `PRAGMA foreign_keys=ON` enabled. It verifies the student delete cascades through enrollments and submissions, confirms the instructor RESTRICT case returns 409, and checks that deleting one student doesn't touch another's data.
2. PR description now matches the code, RESTRICT is called out separately from the CASCADE chain.
3. `delete_user` catches the `IntegrityError` and returns a proper 409 via `ConflictError` instead of a generic 500.
4. `submission.py` now has session-based ownership checks on `get_submissions`, `upload_submission`, `get_latest_submission`, `get_active_submission`, and `activate_submission`, plus a session check on `rerun_submission_autograder`.

Also confirmed the instructor delete dialog now matches the RESTRICT behavior (tells the admin to reassign courses first, button disabled while they have active courses), and `FRONTEND_ORIGIN` logs a warning when unset, which was the other option I gave you.

**New blocking issue: `_verify_student_owner` breaks instructor and TA access to student submissions**

`_verify_student_owner` in `submission.py` only allows access when the session user matches the requested `student_id`, with no allowance for instructors or TAs. But several existing frontend pages call these same endpoints on behalf of other students:

- `manageSubmissions/index.js` loops over the course roster and calls `get_active_submission?student_id=<each student>` to build the manage-submissions table.
- `RegradeRequests.js` calls `get_active_submission?student_id=<requesting student>` for every open regrade request in the instructor queue.
- `submissionHistoryModal.js` (reached from `/assignmentResult/:submissionId`, which both of the above pages navigate to) resolves `studentId` from the submission itself, not the viewer, then calls `get_submissions`, `get_active_submission`, and `activate_submission` with that id.

With the new check, any instructor or TA session hitting these will get a 403 instead of the student's data. That breaks the manage-submissions dashboard, the regrade queue, and the grading/activation flow for course staff. This is worse than what shipped before for anyone using those pages, even though it correctly closes the student-on-student privilege escalation.

The new `test_submission.py` tests won't catch this. The `mock_verify_student_owner` autouse fixture mocks the check out for every test, so nothing exercises it against a non-owner session.

Fix needs a role check alongside the ownership check: allow when the session user is the instructor or TA for the course the student is enrolled in, in addition to the self-match case. `_verify_student` in `code_editor.py` looks fine as-is, I didn't find any instructor call sites for those six endpoints.

Once that's sorted, this should be ready to merge.

DarioDeSilva:
Reviewed this thoroughly, including a cross-check against #330 (feature/ta-role-implementation), which is open in parallel and touches a lot of the same surface area.

## What's solid here

- Cascade delete / RESTRICT modeling is correct: student-owned data cascades, instructor-owned courses are RESTRICT-protected, and the migration's upgrade/downgrade are symmetric with the model changes.
- The new integration tests (`test_cascade_delete_it.py`) exercise real DB-level FK behavior via `PRAGMA foreign_keys=ON` instead of mocking it away, which is exactly right for this kind of change.
- `_verify_student_owner` (self OR instructor/TA of the course) is the right shape for allowing grading access without over-broadening the check.
- The codeEditor/assignments null-date guards fix real crashes, and the frontend delete-confirmation copy now accurately reflects that instructor deletion can be blocked.

## Blocking issues

1. **`delete_user` has no authorization check at all** (`backend/routes/user.py`). Anyone who can reach the API, no login, no session, no role check, can `DELETE /delete_user?id=<any-id>`. This is the exact route this PR modifies, and cascade delete now means a successful call wipes enrollments, submissions, code drafts, extensions, and test results for that user. The blast radius of an already-open hole just got bigger. Needs a permission check before merge.

2. **`create_user` allows self-assigned `role: "admin"` with no auth** (same file). Any anonymous POST can create an admin account. Same file as #1, same theme: every other blueprint touched in this PR got real auth added, user.py got none.

3. **`rerun_submission_autograder` is an IDOR** (`backend/routes/submission.py`). It only checks that *some* user is logged in, not that they own the submission or are course staff, unlike every sibling endpoint in the same file that got `_verify_student_owner`. It also returns the full `SubmissionSchema` dump (`student_code_file`, `results`, `score`) in the response, so any authenticated student can pull another student's submission code and grade by guessing/observing a `submission_id`.

4. **`get_submission_details`, `get_results`, `get_all_assignment_submissions`, `delete_submission` remain fully unauthenticated.** `get_results` takes a plain `email` param and returns that student's submission with no session check at all, anonymous IDOR. `delete_submission` allows anonymous deletion. Given this PR's whole premise is closing exactly this category of gap in `submission.py`, these read as missed rather than deliberately deferred.

## Status of these in #330

I checked #330 directly since it's open in parallel and also does an auth pass:

- `delete_submission` and `rerun_submission_autograder`: **fixed there** (`require_authenticated()` + `require_course_role(...)`, and an owner-or-staff check on the rerun).
- `delete_user`, `create_user`, `get_submission_details`, `get_results`, `get_all_assignment_submissions`: **not touched in #330 either**, same gaps.
- Separately: #330 branches off `main` before this PR and does not include these `_verify_student_owner` additions at all, so in #330's copy of `submission.py`, `get_submissions`, `upload_submission`, `get_latest_submission`, `get_active_submission`, and `activate_submission` currently have **no auth**. It also has its own independent session-login implementation. Whichever of these two PRs merges second is going to need a real rebase/reconciliation in `routes/user.py` and `routes/submission.py`, not just a mechanical one.

## Suggested path

Given #330 already built general-purpose auth helpers (`require_authenticated`, `require_course_role`, proper 401/403 semantics) rather than the one-off `_verify_student_owner`/`_verify_student` here, it'd be worth landing #330 first and rebasing this PR on top of it to reuse those helpers instead of the bespoke versions. That also removes the duplicate-implementation risk from the two PRs solving the same problem independently. While one of these two PRs is being finished, could we also close `delete_user` and `create_user` specifically, since those are the two most severe holes and don't require the DB/session work in this PR to fix, just a role check?

## Non-blocking notes

- Migration `c4d5e6f7a8b9`'s docstring header says "Revises: b2c3d4e5f6a7" but `down_revision` is actually `1bdb41066778` (the real merge-heads point, chain resolves fine). Stale comment, worth a one-line fix so migration history reads correctly later.
- `_check_run_code_rate_limit` is a per-process in-memory dict: under multiple workers the effective limit becomes `10 x num_workers`, and a caller who never returns leaves a stale entry indefinitely since eviction only happens on their next call. Not urgent, just won't behave exactly as documented in a real multi-worker deployment.
- `test_code_editor.py::test_submit_code_past_due_date` sets `mock_assignment.enable_code_editor = True` twice, copy-paste leftover, harmless.

## PR size

At +835/-603 across 21 files, this is large for its stated scope (cascade delete on user deletion). It bundles in CORS locking, mandatory `SECRET_KEY`, and a full session-auth + rate-limiting pass over `code_editor.py` and `submission.py`, plus a couple of unrelated frontend null-date fixes. Worth splitting auth hardening into its own PR in the future so each piece can be reviewed and reverted independently, though I recognize this one is already up and functional.

DarioDeSilva (approved):
Re-reviewed after the latest two commits. All the blocking issues from my last pass are fixed:

- `delete_user` now requires an authenticated admin session.
- `create_user` still allows unauthenticated student self-signup, but creating admin/instructor accounts now requires an admin session.
- `rerun_submission_autograder` now requires course staff or admin (previously any logged-in user could rerun and get back the full submission dump).
- `get_submission_details`, `get_results`, `get_all_assignment_submissions`, and `delete_submission` are all now properly authorized.

Tests are updated consistently with the new checks, and I ran the full backend suite against the latest commit: 296 passed, 0 failed.

A couple of very minor non-blocking nits, not worth holding this up for:
- `rerun_submission_autograder` checks the session twice (once directly, once inside `_verify_course_staff`), harmless redundancy.
- `get_submission_details` and `delete_submission` fetch the record before checking authorization, so an unauthenticated caller gets a 404 vs 403 depending on whether the id exists. Low severity since these are UUIDs.

Nice work closing these out. Approving.

kiat:
Thanks!

PR #317, opened 2026-06-20
Deactivative/Reactivate Courses

rosered123:
Added a deactivate/reactivate button in the course settings. 

Deactivating a course allows the course to still be present in the teacher's dashboard so they can refer back to it or use its data, but it removes it from a student's dashboard. Reactivating the course brings it back to the student's dashboard.


SashankN7:
Could you fix the merge conflicts before I review?

SashankN7 (changes requested):
**Critical Issues**
1. Missing backend endpoint — feature is broken
The frontend calls  set_course_active  via  PUT /set_course_active , but  backend/routes/course.py  has zero changes in this diff. The endpoint doesn't exist. Clicking "Deactivate Course" will 404. This feature literally doesn't work.
2. No migration for the  active  column on courses
 models.py  adds  active = db.Column(db.Boolean, default=True, nullable=False)  to  Course , but the only migration ( f3a8b2c1d4e5_add_password_reset_to_users.py ) adds  reset_token  and  reset_token_expiry  to  users . There is no migration for the  active  column on  courses . The app will crash with a missing column error at runtime.
3. Two unrelated features in one PR
"Forgot Password" and "Deactivate/Reactivate Courses" are completely separate features with no shared logic. These should be two separate PRs — they have different review scopes, different risk profiles, and different test requirements.

**Security Issues**
4. Reset token stored in plaintext
 reset_token  is stored as a plain  String  in the database. If the DB is compromised, all active reset tokens are immediately usable. Should be hashed (same as passwords — use  hash_password() / verify_password()  that already exist in the codebase).
5. Dev fallback leaks reset links to logs
// python
if not email_address or not email_password:
    print(f"[DEV] Password reset link for {to_email}: {reset_link}", flush=True)
If someone deploys without setting  EMAIL_ADDRESS / EMAIL_PASSWORD , every reset link gets printed to stdout/logs. In a shared hosting or containerized environment, this is an information leak. Should use  logger.warning()  at minimum, or better — raise an error in production.
6. No rate limiting on  /forgot_password 
An attacker can POST to  /forgot_password  in a loop to spam a user's email (email bombing) or to enumerate which emails have accounts (despite the generic response, timing differences between "user exists" and "user doesn't exist" paths leak this info).
7. Potential TypeError in  reset_password 
// python
if db_user.reset_token_expiry < datetime.now(timezone.utc):
If  reset_token_expiry  is  None  (edge case), this throws  TypeError: '<' not supported between instances of 'NoneType' and 'datetime' . Should check  if not db_user.reset_token_expiry or db_user.reset_token_expiry < ... .

**Code Quality Issues**
8. Duplicate import in App.js 
// javascript
import AdminCourseAdd from "./pages/admin/courses/add"; // <-- Add this line
import AdminCourseAdd from "./pages/admin/courses/add";
The same import appears twice with a stray comment. This will cause a linting error.
9.  active  column default but no DB default
The model has  default=True  (Python-side default), but the migration for courses is missing entirely. Even if you add the migration, existing courses need a sensible default —  default=True  in the model is fine for new rows, but the migration  ALTER TABLE  should also set a default for existing rows.
10.  _send_reset_email  has no error handling for SMTP failures
If  smtplib.SMTP  raises (e.g., network timeout, auth failure), the exception propagates unhandled. The caller catches it, but the error message just says "Failed to send reset email" without distinguishing transient vs. permanent failures.
11. Unused imports
 requests ,  random ,  string  are imported in  user.py  but not used by the new code (possibly pre-existing).
12. Missing  down_revision  chain verification
The migration  f3a8b2c1d4e5  has  down_revision = '92009ca5c92a' . Need to verify this is the current head of the migration chain — if another migration was added to  main  after this branch was created, the chain is broken.

**What's Good**
- The forgot password UI ( forgotPasswordModal.js ,  resetPassword/index.js ) is clean and well-structured
- The email enumeration protection (returning the same message regardless of whether the email exists) is correct
- The token expiry logic (1-hour window, token cleared on use) is sound
- The course deactivation UI (Popconfirm pattern, optimistic state update with rollback) follows the project's existing patterns
- The  active  column design (keep course in DB, filter from student view) is the right approach

kiat:
Please fix the merge conflicts again!

rosered123:
merge conflicts are fixed!

kiat:
Please update it again!

PR #318, opened 2026-06-21
Fix late submission settings persistence and validation

### LINKED ISSUE #313: Student cannot submit an assignment passed the due date WITH late submission enable and late due date set
When instructor enable late submission and set the late due date in an assignment setting, late due date is updated on the student side, as well as under extension from the instructor side, however, student cannot submit an assignment that is late, an attempt to submit just going to received "due date has passed" 

Image

The image above enable late submission and set late due date

Image

Image

Student being able to see the late due date, and in extension from instructor side late due date also present. But if student trying to submit it late, then it show due date has passed even though it still far away from late due date.

keiiii3257:
Solve issue #313 
Problem: The late submission checkbox and late due date never save after enabling. The student is able to see the late due date, but cannot submit it late within the configured late window,  and the late due date is still able to be set and seen by the student with late submission disabled. 

Frontend
- Load and save allowLateSubmissions correctly in Assignment Settings.
-  Allow assignment access during the valid late submission window.
- Clear late_due_date when late submissions are disabled.
- Disable the late due date picker when late submissions are not enabled.

Testing
- Verified the persistence of enabling "allow late submission" and the late due date state after reopening the assignment setting.
- Verified that students can submit after the original due date when late submissions are enabled and the current time is before the late due date.
- Verified that disabling “Allow Late Submissions” clears the late due date after saving.
- Verified that students cannot submit late when late submissions are disabled.
- Verified that the late due date picker is disabled when late submissions are not enabled.

Improvement (I haven't done this, but it could be added on) 
- Add a late due column for the instructor in the dashboard, just like the student
- While the assignment has a late submission disabled, create a late extension for a specific student, allowing them to submit it late (right now, still able to set a late due date extension, it's just because the assignment setting itself has a late submission disabled; the student with an extension is still not able to submit late)
- Add a column showing whether that student submitted the assignment late or not (inside the manage submission and review grades tab) 
- On the student side, they will be able to see whether they submit that assignment late with a "late" label


yczhangcc (approved):
The late submission setting now works correctly, and I also tested the extension flow for a specific student. Both worked as expected. 
This is a smart solution and a useful feature improvement. Thanks for thinking through both the general late submission case and the student-specific extension case. Nice job!

PR #322, opened 2026-06-24
Add Ollama (Local LLM) support for AI feedback

Gaurav-021:
# Pull Request: Add Ollama (Local LLM) Support for AI Feedback

## Overview
This PR adds support for **Ollama** as a local LLM provider for generating structured student feedback, alongside ChatGPT, Gemini, and Claude. This allows instructors to host their own open-source models (like `llama3`, `deepseek-coder`, etc.) locally or on private servers.

## Key Changes
* **Database:** Added `ollama_base_url` column to the `courses` table (migration script included).
* **Backend:** Integrated Ollama chat API generation using `"format": "json"` to enforce valid JSON responses. Added model tags fetching and connection testing routes.
* **UI:** Updated Course AI Settings, Assignment Settings, and Create Assignment views to include Ollama. Custom-tailored settings view to take plaintext URLs instead of masked API keys.
* **Tests:** Added unit tests verifying Ollama connection test success. All tests pass.

---

## Setup Note: Docker & Local Ollama Connection
If running this application locally using Docker Compose, the backend container routes requests back to your host machine via `http://host.docker.internal:11434`. 

To allow this connection, you must configure Ollama on your host machine to bind to `0.0.0.0` (accepting connections outside localhost). 
Please follow the official instructions in the [Ollama FAQ Guide](https://github.com/ollama/ollama/blob/main/docs/faq.md) for your OS (Linux systemd, macOS `launchctl`, or Windows Environment Variables) to set `OLLAMA_HOST=0.0.0.0`.


DarioDeSilva (changes requested):
**Duplicate revision ID breaks migrations** (`backend/migrations/versions/`)
Two migration files share the same revision ID `a1b2c3d4e5f6` — `a1b2c3d4e5f6_add_ai_settings_fields.py` and `a1b2c3d4e5f6_add_code_drafts_table.py`. Alembic warns `Revision a1b2c3d4e5f6 is present more than once` and cannot reliably determine the migration chain. Running `flask db upgrade` on a clean database will apply one arbitrarily and skip the other, leaving the schema incomplete.

**SSRF via unvalidated `ollama_base_url`** (`backend/routes/course.py:677, 732, 943, 1146`)
`ollama_base_url` is saved directly from instructor input with no URL validation, then used verbatim in `requests.get(f"{base_url}/api/tags")` and `requests.post(f"{base_url}/api/chat")` across three endpoints. An instructor could set the URL to `http://169.254.169.254/latest/meta-data/` or any internal service, and the backend will make that request on every model test or AI feedback call — a full SSRF into the Docker network or cloud metadata service.

**Default URL makes Ollama appear "Connected" on every course** (`backend/api/models.py:37`, `backend/routes/course.py:591`)
`default="http://host.docker.internal:11434"` is non-empty, so `bool(course_obj.ollama_base_url)` is `True` for every course after migration. This means:
- `has_ollama_api_key` is always `true` in the API response
- The "AI configured" badge in Assignment Settings and Create Assignment is permanently green, even for courses with no AI provider set up
- The AI Settings page shows "Connected" for every course — the "Using Default" status branch can never fire

Fix: change the model default to `""`. The runtime fallback `(api_key or "http://host.docker.internal:11434")` already exists in all three route handlers and handles the empty case correctly.

**`ollama_base_url` exposed in plaintext in the API** (`backend/routes/course.py:592`)
Every other provider returns only `has_X_api_key: bool`, never the actual credential. Ollama returns the raw URL directly. If an instructor configured a private internal endpoint, that address is leaked to any API consumer.

**Exception details returned to client** (`backend/routes/course.py:740, 954, 1180`)
All three Ollama error handlers return `str(e)` or `response.text` directly in the JSON response — e.g. `"Failed to connect to Ollama at {base_url}: {str(e)}"`. A connection error exposes internal hostnames, port numbers, and socket details. Log these server-side and return a generic error message instead.

**New Ollama test only covers the request-body path** (`backend/test/unit/test_course.py:663`)
The test provides `api_key` directly in the request body, which bypasses the branch that reads `course_obj.ollama_base_url` from the DB. A regression in the DB lookup path or the `BadRequestError` fallback for a missing saved URL wouldn't be caught.

SashankN7 (changes requested):
Gaurav's commits fixed 6 out of the 7 issues Dario pointed out in his review. However, the SSRF issue is the main thing that still needs to be addressed before merging.

**1. SSRF — Still not addressed (critical)**
None of the 7 commits add URL validation. The  ollama_base_url  is still stored and used directly in  requests.get()  /  requests.post()  with no validation. An instructor could set  http://169.254.169.254/latest/meta-data/  and the backend will happily make that request. This is a blocking security issue.

**2. Test for default URL fallback is backwards**
Commit  4ffd46b  adds  test_test_ai_model_ollama_missing_saved_url_error  which asserts a 400 error when  ollama_base_url="" . But the runtime code does:
// python
base_url = (api_key or "http://host.docker.internal:11434").strip()
An empty string is falsy in Python, so  "" or "http://host.docker.internal:11434"  evaluates to the default URL. This means:
- The test expects a 400 error for empty URL
- But the code actually falls back to the default URL and makes the request
These contradict each other. Either the test is wrong (it should mock a successful request to the default URL) or the code should actually error when no URL is configured. Need to clarify the intended behavior.

**3. Minor code duplication**
The Ollama blocks in  fetch_ai_models ,  test_ai_api_key , and  test_ai_model  each duplicate ~30-45 lines of similar request/response logic. Could be extracted into a shared helper, but this is a style improvement, not a blocker.

Gaurav-021:
Everything should now be addressed including a solution for the security issue.

SashankN7 (changes requested):
Overall, the PR looks good. However, #1 and #2 below need to be addressed before this can merge. The other four are less complex nice-to-haves.

**1. Migration uses raw SQL — breaks SQLite (used in tests)**
 backend/migrations/versions/d4e5f6a7b8c9_add_ollama_base_url.py 
The  upgrade()  uses Postgres-specific raw SQL:
// python
op.execute("ALTER TABLE courses ADD COLUMN IF NOT EXISTS ollama_base_url VARCHAR DEFAULT ''")
 ADD COLUMN IF NOT EXISTS  is Postgres-only syntax. If the test suite ever uses SQLite (which many Flask test setups do), this migration will fail. Use the Alembic API instead:
// python
op.add_column('courses', sa.Column('ollama_base_url', sa.String(), server_default=''))
Same issue in  downgrade()  with  DROP COLUMN IF EXISTS .

**2. Migration down_revision will conflict with other open PRs**
 backend/migrations/versions/d4e5f6a7b8c9_add_ollama_base_url.py 
The migration chains from  down_revision = 'b2c3d4e5f6a7' . PR #312 (cascade delete) and PR #330 (TA role) also chain from that same revision. When all three are merged, this creates multiple Alembic heads. The migration should be rebased onto whichever PR merges first so the chain is linear.

**3. get_provider_credentials returns URL as api_key — misleading variable name**
 backend/ai_feedback/integration.py  —  get_provider_credentials() 
For Ollama, the function returns the base URL string stored in the variable named  api_key :
// python
elif provider == "ollama":
    api_key = (course.ollama_base_url or "http://host.docker.internal:11434").strip()
This  api_key  is then passed to  get_structured_feedback_from_ollama()  as  base_url . The naming is confusing — a reader expects an API key but gets a URL. Consider renaming to  credential  or  connection_string , or at minimum add a comment clarifying that for Ollama this is the base URL, not an API key.

**4. Double validation of Ollama URL**
Files:  backend/ai_feedback/integration.py  +  backend/routes/course.py 
 validate_ollama_url()  is called in three places:
1.  get_structured_feedback_from_ollama()  in  integration.py 
2.  _request_ollama()  in  course.py 
3.  update_ai_settings()  in  course.py 
The integration.py function validates the URL before making the API call, but  _request_ollama()  also validates it. This means the URL is validated twice per request. Consider validating once at the storage layer ( update_ai_settings ) and trusting it downstream, or only in the integration layer — not both.

**5. Missing test: fetch_ai_models for Ollama**
There are tests for  test_ai_model  and  test_ai_api_key  for Ollama, but no test for the  fetch_ai_models  endpoint with  provider=ollama . The Ollama branch in  fetch_ai_models  parses models from the  /api/tags  response — worth a quick test to ensure the model name extraction works correctly.

**6. Minor: ollama_base_url model default vs migration server_default**
 backend/api/models.py 
The column is defined as  default=""  (Python-side default), but the migration sets  server_default='' . These are fine for Postgres, but the model and migration should be consistent. If the column should never be NULL, the migration should also specify NOT NULL .

SashankN7 (changes requested):
**Critical / Must Fix**

**Comment 1** —  backend/util/url_utils.py  line 30:  urllib.parse.urlparse  doesn't raise on invalid input
urllib.parse.urlparse() never raises an exception — it silently parses garbage
into an object with empty fields. The `try/except Exception` block is dead code
and will never execute. For example:
 
  urlparse("not-a-url")  →  ParseResult(scheme='', netloc='', path='not-a-url', ...)
 
The hostname check below will catch the case where hostname is None/empty, but
this try/except gives a false sense of security. Consider removing it or using
a more explicit validation (e.g. `if not parsed.scheme or not parsed.netloc`).

**Comment 2** —  backend/routes/course.py   _request_ollama : Fragile string-matching for log prefix
// python
log_prefix = "Failed to connect to Ollama" if "Failed" in error_msg_500 else "Ollama connection error"
This derives the log prefix by checking if "Failed" is a substring of the
caller-provided error_msg_500 string. This is extremely fragile — if anyone
changes the error message wording (e.g. "Unable to connect"), this silently
falls through to the wrong branch. The log prefix should be independent of
the user-facing error message. Consider just using a single log prefix like
"Ollama connection error" regardless of the input message, or pass the log
prefix as a separate parameter.

**Comment 3** —  backend/routes/course.py   _request_ollama : Similar string-matching for error logging (lines 147-152)
// python
if "test failed" in error_msg_400:
    current_app.logger.error(f"Ollama connection test failed: {response.text}")
elif "cannot be used" in error_msg_400:
    current_app.logger.error(f"Selected Ollama model cannot be used: {response.text}")
Same concern as above — the log messages are derived by matching substrings
of the error_msg_400 parameter. This couples the logging logic to the exact
wording of caller-supplied strings. If someone changes "test failed" to
"connection test failed" or "cannot be used" to "is not usable", the log
message silently changes too. Consider just logging the raw response with a
generic prefix like "Ollama API error" and let the error_msg_400 be purely
user-facing.

**Comment 4** — Hardcoded default URL  "http://host.docker.internal:11434"  is scattered in 3+ places
The default fallback URL "http://host.docker.internal:11434" appears in:
 
  • integration.py:579 — get_provider_credentials()
  • routes/course.py:130 — _request_ollama()
  • tests — expected values
 
This should be a single constant (e.g. DEFAULT_OLLAMA_BASE_URL) defined
once and referenced everywhere. If the default ever changes (e.g. to
support a different port or host), you'd have to hunt down every copy.

**Comment 5** —  backend/ai_feedback/integration.py : Unused import  BadRequestError 
`BadRequestError` is imported but never used in this file. The
validate_ollama_url() function raises BadRequestError internally, but
the caller in get_structured_feedback_from_ollama doesn't catch it —
it propagates up to the background task's generic Exception handler.
 
If the intent is to catch it specifically, add a try/except. Otherwise,
remove the unused import.

**Suggestions / Improvements**

**Comment 6** —  has_ollama_api_key  is semantically misleading in  routes/course.py  line 683
// python
"has_ollama_api_key": bool(course_obj.ollama_base_url),
Ollama doesn't use API keys — it uses a base URL. Calling the field
"has_ollama_api_key" is misleading to anyone reading the API response
or the frontend code. Consider renaming to "has_ollama_url" or
"has_ollama_configured" for clarity. This would need a coordinated
rename in the frontend files (assignmentSettings, aiSettings,
CreateAssignment) as well.

**Comment 7** —  routes/course.py   fetch_ai_models : Missing  validate_ollama_url  before API call when URL is user-supplied
// python
# Line ~819: When api_key is passed directly (not from DB), the URL isn't validated
if provider == "ollama":
    api_key = course_obj.ollama_base_url   # ← OK, from DB
# But when api_key is user-supplied via request body:
if provider == "ollama":
    response, err_response = _request_ollama(api_key, ...)  # ← _request_ollama does validate
The `fetch_ai_models` endpoint does call _request_ollama which includes
validation, so this is OK at runtime. However, in test_ai_model and
test_ai_api_key, when `api_key` is passed directly in the request body,
there's no explicit validate_ollama_url() call before using it in the
test flow. The _request_ollama helper does validate, but the validation
happens deep inside the call chain. Consider adding a single validation
gate at the top of each endpoint (after the provider is known) so the
SSRF check is clearly visible and doesn't depend on the helper doing it.

**Comment 8** —  backend/routes/course.py   test_ai_model : No test for Ollama connection failure / timeout scenarios
The existing tests cover the happy path (success, saved URL, fallback
URL) but there's no test for:
  • Connection refused (Ollama server not running)
  • Timeout
  • Non-200 response from Ollama
  • Invalid/unparseable response
 
These are realistic failure modes for a self-hosted service. Consider
adding at least a connection-refused and a timeout test case, since the
error handling paths through _request_ollama are non-trivial.

**Comment 9** —  backend/routes/course.py : Inconsistent use of  _request_ollama  vs direct  requests.get/post 
The Ollama endpoints use the `_request_ollama` helper, which is good.
But in `get_structured_feedback_from_ollama` (integration.py), the Ollama
call is made with raw `requests.post()` without using _request_ollama.
 
This means the URL validation and error handling patterns differ between
the route-level test endpoints and the actual feedback generation path.
Consider having integration.py also use a shared helper, or at minimum
ensure the validation/error handling is consistent.

**Comment 10** —  backend/api/models.py :  server_default  on the Column vs raw SQL in migration
// python
# models.py
ollama_base_url = db.Column(db.String, default="", server_default="")
// python
# migration
op.execute("ALTER TABLE courses ADD COLUMN IF NOT EXISTS ollama_base_url VARCHAR DEFAULT ''")
The model declares `server_default=""` which is good, and the migration
uses raw ALTER TABLE with DEFAULT '' which matches. However, the
migration downgrade uses raw `op.execute` instead of
`op.drop_column('courses', 'ollama_base_url')` — inconsistent with
the upgrade pattern. Consider using op.add_column / op.drop_column
for consistency with other migrations in the project, and add `sa.Column`
with the proper type.

kiat:
Please fix the conflict so that we can merge it asap.

SashankN7 (changes requested):
**Blocking/Critical**
1.  backend/util/auth.py  was DELETED — will break main on merge
PR #312 added session-based auth via  require_authenticated()  and  require_course_role() . This PR deletes that file entirely, along with  test_auth_util.py  and  test_app_config.py . If merged, main loses its auth infrastructure and test coverage. This is outside the scope of "Ollama support" and must be removed from this PR.
2. 401 interceptor removed from  frontend/src/services/index.js 
PR #312/330 added a 401 response interceptor that clears  userInfo ,  courseInfo ,  courseRole  from localStorage and redirects to  / . This PR removes it entirely. Expired sessions will no longer auto-logout users — they'll get silent failures instead.
3. Model PK type mismatch —  AIFeedbackRequest  and  AIChatMessage  use  db.String  for PKs
Every other model uses  UUID(as_uuid=False) . These two new models use  db.String  for  id ,  student_id , and  assignment_id . The foreign keys reference  users.id  and  assignments.id  which are UUIDs on main — this will cause type mismatches at the database level. Fix:
// python
id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id", ondelete="CASCADE"), nullable=False)
assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id", ondelete="CASCADE"), nullable=False)
4. Missing migration for  ai_chat_messages  table
The  AIChatMessage  model exists but no migration file was found for it on the branch. Only  a2b3c4d5e6f7_add_ai_feedback_requests_table.py  exists. Without a migration, the table won't be created in production.
5.  config.py  lost all  SESSION_COOKIE_*  settings
Main's  config.py  has  SESSION_COOKIE_HTTPONLY ,  SESSION_COOKIE_SAMESITE , and  SESSION_COOKIE_SECURE  logic from PR #330. This PR's  config.py  is a bare skeleton — just  SQLALCHEMY_DATABASE_URI  and  TestConfig . The session cookie hardening is gone.
**Medium**
6. No timeout on  test_ai_model  Ollama call
The  test_ai_model  endpoint calls  _request_ollama()  which makes an HTTP request to the Ollama server. If Ollama is slow or unresponsive, this endpoint will hang indefinitely. Add a timeout:
// python
response = requests.post(f"{ollama_url}/api/chat", json=..., timeout=30)
7. DNS rebinding SSRF risk in  validate_ollama_url 
The whitelist approach ( localhost ,  host.docker.internal ,  127.0.0.1 ) is good, but doesn't prevent DNS rebinding. An attacker could set up a DNS record for a whitelisted hostname that resolves to  169.254.169.254  (cloud metadata endpoint). Consider resolving the hostname to an IP and checking it against private ranges before making the request, or document this as a known limitation.
8. Error messages leak  base_url  to the frontend
In  get_structured_feedback_from_ollama :
// python
raise ValueError(f"Failed to connect to Ollama: {str(e)}")
raise ValueError(f"Ollama API error ({response.status_code}): {response.text}")
These messages include the internal Ollama URL. Sanitize or genericize error messages in production.
9.  test_ai_model  doesn't validate  model  parameter for Ollama
When provider is  ollama , the endpoint accepts any  model  string without validating it exists on the Ollama server. A typo produces a confusing error from the Ollama API rather than a clear validation error.
**Minor**
10.  AIChatPanel.test.jsx  missing error path tests
No test for when  getAssignmentPrompts  fails (network error, 404). The component should gracefully handle this, and the test should verify it.
11. Chat history truncation is invisible
 get_chat_history  limits to 20 messages, but the UI doesn't indicate that older messages are hidden. A "Showing last 20 messages" notice would improve UX.
12. Commit history should be squashed
The branch has 8 commits with messages like "refactor: address PR feedback" and "fix: clean up URL validation". These suggest iterative development that should be squashed before merge.

SashankN7 (changes requested):
**Critical / Must Fix**
1. Existing migrations modified — will break already-deployed databases
> You changed the  down_revision  in two existing migrations ( c4d5e6f7a8b9  and  d4e5f6a7b8c9 ). If these migrations have already been applied to any database (dev, staging, prod), Alembic will see the changed  down_revision  and attempt to re-run them, causing errors or data corruption. Existing migrations should never be modified after they've been applied. Instead, create new migrations that sit at the end of the chain and adjust the  down_revision  of only your new migrations to point to the current head.
2.  get_structured_feedback_from_ollama  swallows original exception
> The  except Exception as e: raise ValueError("Failed to connect to Ollama server")  discards the original exception's type, message, and traceback. This makes debugging connection failures extremely difficult. Use exception chaining:
> // python
> except Exception as e:
>     raise ValueError(f"Failed to connect to Ollama server: {e}") from e
> Or better yet, don't catch it here — let the caller handle connection errors with appropriate context.
3. Security warning removed from  __init__.py  — unrelated to this PR
> The  SESSION_COOKIE_SECURE  warning was removed from  api/__init__.py . This warning alerts developers when session cookies are being sent over plain HTTP — a real security risk. Removing it reduces visibility into misconfigurations. This change is unrelated to Ollama support and should be reverted (or moved to a separate PR with justification).
4. README documentation removed — unrelated to this PR
> The README section documenting  SECRET_KEY ,  FRONTEND_URL ,  SESSION_COOKIE_SAMESITE , and  SESSION_COOKIE_SECURE  env vars was deleted. This is developer-facing documentation that helps new contributors set up the project. Removing it reduces onboarding clarity. This is unrelated to Ollama support and should be reverted.
5. Race condition in  check_feedback_limits  — concurrent requests can exceed max
>  check_feedback_limits  counts existing requests and checks the limit, but  record_feedback_request  is called separately. Two concurrent requests could both pass the limit check and both record, exceeding  ai_feedback_max_requests . Use a database-level constraint or a  SELECT ... FOR UPDATE  lock:
> // python
> with db.session.begin_nested():
>     # count + check + record in a single transaction

**Important / Should Fix**
6. Inline imports in  settings.py  — potential circular import issues
>  check_feedback_limits ,  record_feedback_request ,  get_student_feedback_status ,  get_chat_history , and  store_chat_message  all have imports inside the function body ( from api.models import ... ,  from api import db ). This pattern is unusual and can mask circular import issues at module load time. If these imports are at the top of the file, any circular dependency would fail fast. Move them to module-level unless there's a specific reason they're inline.
7.  has_ollama_api_key  renamed to  has_ollama_configured  — verify all references
> The frontend field was renamed from  has_ollama_api_key  to  has_ollama_configured  in three places. Make sure the backend  get_course_info  route also returns  has_ollama_configured  (not  has_ollama_api_key ), otherwise the frontend will always see  undefined  and the Ollama status indicator will never show success.
8.  onSendMessage  contract changed — breaking API change
>  AIChatPanel 's  onSendMessage  callback now expects a  { reply, feedback_status }  object instead of a plain string. All consumers need to be updated. Verify that every parent component passing  onSendMessage  handles the new return shape, otherwise the chat will show  [object Object]  or fail silently.
9. No URL validation hardening for Ollama base URLs — SSRF risk
> The Ollama  base_url  is user-provided and passed directly to  requests.post() . An attacker could set it to  http://169.254.169.254  (cloud metadata),  http://localhost:6379  (Redis), or other internal services.  validate_ollama_url  should reject private/internal IPs and link-local addresses, not just check the URL format.
10.  get_chat_history  has hardcoded  limit=20  — should be configurable or documented
> The default  limit=20  is arbitrary. If an instructor wants students to have longer conversation history, there's no way to configure it. At minimum, add a comment explaining why 20 was chosen. Better: make it a parameter or tied to a setting.
11.  record_feedback_request  doesn't validate  student_id / assignment_id  exist
> The function accepts raw IDs and inserts them directly. If the IDs don't exist (e.g., due to a bug or stale frontend state), the FK constraint will raise a raw  IntegrityError  instead of a user-friendly error. Consider validating existence first or catching the specific exception.
12. Frontend  AIChatPanel  still imports unused  DEFAULT_PROMPTS  constant
> The component was refactored to fetch prompts from the backend, but the old  DEFAULT_PROMPTS  import from  aiFeedbackSettings.js  may still be referenced. Clean up any dead imports to avoid confusion.

**Minor / Nice to Have**
13.  countdown > 0  as useEffect dependency — works but fragile
>  useEffect(() => { ... }, [countdown > 0])  uses a boolean expression as a dependency. This intentionally only re-runs when countdown crosses zero. It works, but it's a non-obvious pattern that could confuse future contributors. Add a brief comment explaining the intent.
14. Trailing whitespace in  api/__init__.py 
> Several lines have trailing whitespace added (e.g., blank lines between code blocks). Minor style issue but linters will flag it.
15.  canSubmitOnTime  and  canSubmitLate  variables are redundant
> In  assignments/index.js ,  const canSubmitOnTime = !dueDateHasPassed  and  const canSubmitLate = inLateWindow  are defined but just alias existing variables. The original code used  dueDateHasPassed  and  inLateWindow  directly in the navigate condition. These new variables add indirection without improving clarity.
16.  .gitignore  removing  frontend/package-lock.json 
>  package-lock.json  should generally be committed for reproducible builds. If there's a reason to exclude it (e.g., it's huge and causes merge conflicts), that should be documented.
17. Test coverage for  check_feedback_limits  edge cases
>  test_ai_feedback_request_tracking.py  covers basic cases, but doesn't test the race condition scenario, concurrent requests, or what happens when  max_requests=0  (disabled). Consider adding edge case tests.
18.  get_student_feedback_status  and  check_feedback_limits  duplicate query logic
> Both functions query  AIFeedbackRequest  count and check wait times. Extract the common query into a shared helper to reduce duplication and keep the logic in sync.

PR #326, opened 2026-06-26
Fix assignment visibility and scheduled publishing

### LINKED ISSUE #314: Unpublish an assignment but student still be able to see it on their side
When an instructor create an assignment, and then later set that assignment to unpublish, on the instructor side in the image below it shown assignment is not publish 

Image

But then student still able to see that assignment 

Image

keiiii3257:
close #314 
Instructor are able to to set publish/unpublish state of an assignment. 

Expected behavior:
Instructor: 
- The table in the dashboard and assignment tab should show all assignment
- Instructor dashboard assignment count all assignments 
Screenshot 2026-07-14 at 5 58 20 PM

- Able to set the publish/unpublish state of an assignment
- A new indicator of publishing status for assignments in the dashboard and assignment tab's table

Student:
- Student cannot see unpublished assignments.
- Student dashboard assignment count only visible assignments.

Frontend
- Hide unpublished assignments from student assignment list
- Hide scheduled assignments from students until their release date
- Count only visible assignments on the student dashboard
- Update instructor publish indicator to reflect scheduled release state

Backend
- Require `user_id` when requesting course assignments
- Validate that the requesting user is enrolled in the course
- Return assignments based on the requester's course role
- Instructors/TAs receive all assignments
- Students only receive published assignments 
- Return `400` for missing `user_id`
- Return `403` for users not enrolled in the requested course

Limitations / Future Enhancement (Separate PR)
- Scheduled assignments are still displayed as **Unpublished** instead of **Scheduled** in:
  - `frontend/src/pages/instructor/dashboard/index.js`
  - `frontend/src/pages/instructor/assignments/Assignments.js`
- Scheduled assignments are not automatically published once their release time is reached. Automatic release scheduling is outside the scope of this PR.


frontend/src/pages/assignments/index.js
```diff
@@ -93,9 +93,17 @@ export default function Assignments() {
           new URLSearchParams({ course_id: urlParams.courseId })
         );
         const assignmentsData = await assignmentsResponse.json();
-
+        
+        const now = moment();
+        const visibleAssignments = assignmentsData.filter((assignment) => {
+          const releaseDate = assignment.published_date
+            ? moment(assignment.published_date)
```
DarioDeSilva: This condition hides any published assignment that has no release date. `releaseDate` is `null` when `published_date` is not set, and `null` is falsy, so the `&&` short-circuits to `false`. That means every assignment published without a scheduled date -- including all existing ones -- is invisible to students. The condition should be:

```js
return assignment.published && (!releaseDate || !now.isBefore(releaseDate));
```

Same bug on the identical line in `dashboard/index.js`.

frontend/src/pages/instructor/assignments/CreateAssignment.js
```diff
@@ -158,7 +158,7 @@ export default ({
         name: values.name,
         course_id: courseId,
         published_date: toIso(values.releaseDate),
-        published: releaseDate ? releaseDate <= now : false,
+        published: true,
```
DarioDeSilva: The old value was `published: releaseDate ? releaseDate <= now : false`, which kept scheduled assignments unpublished in the DB until their release date. Always setting `true` here means the backend's `assignment.published` check is bypassed -- `submit_code` and `run_code` both gate on that field server-side. A student who knows an assignment ID can POST directly to the API before the release date and the server will accept it. The scheduling enforcement needs to live server-side, not only in the frontend filter.

frontend/src/pages/instructor/assignments/Assignments.js
```diff
@@ -6,7 +6,15 @@ import { useState } from "react";
 import { Link, useParams } from "react-router-dom";
 import { getCourseAssignments } from "../../../services/course";
 // import { tableData } from "./constant";
-
+const getPublishStatus = (record) => {
```
DarioDeSilva: This function is defined identically in `instructor/dashboard/index.js`. Worth moving to a shared util so a future logic change only needs to happen in one place.

DarioDeSilva (changes requested):
Two issues need fixing before this can merge, plus one minor cleanup note.

frontend/src/pages/instructor/assignments/Assignments.js
keiiii3257: Hello, thanks for reviewing. I have updated the changes and move the function to /codeAssist/frontend/src/common/assignmentVisibility.js

frontend/src/pages/instructor/assignments/CreateAssignment.js
keiiii3257: Hello, thanks for reviewing. This is something I definitely did not think about much. I have updated the changes. And due to this, I will not include scheduling to publish an assignment in this PR, i will make an different issues for just scheduling publish and work on it in a different PR.

frontend/src/pages/assignments/index.js
keiiii3257: Hello, thanks for reviewing. While assignments created through the current UI always have a release date (an instructor cannot create an assignment with the release date field not set), this is still a valid edge case that needs to be addressed. You are right with the condition, and I have changed it to make the code more robust.

DarioDeSilva:
## Review

Nice work on the instructor status indicator and the shared `assignmentVisibility` util — a couple of things I think are worth addressing before merging, in particular the backend gap below.

### Critical: filtering happens only on the frontend — the backend still returns unpublished/scheduled assignments to everyone

`GET /get_course_assignments` (`backend/routes/course.py:645-660`) queries and returns every assignment for a course, with no filtering by publish status and no auth/enrollment check:

```python
assignments = db.session.query(Assignment).filter_by(course_id=course_id).all()
assignments = AssignmentSchema().dump(assignments, many=True)
return jsonify(assignments), 200
```

This PR filters the response after it's already fetched into the React state (`assignments/index.js`, `dashboard/index.js`). That means:

- Any client that calls the endpoint directly (devtools network tab, curl, etc.) with a known `course_id` can still see unpublished/scheduled assignment names, due dates, etc.
- This is the actual bug described in #314 — the data is "still visible to students," just not in this one UI surface anymore.

Suggestion: filter by publish status (and check the requester's role/enrollment) server-side in `get_course_assignments`, and keep the frontend filtering as a UX nicety layered on top rather than the only control.

### `isAssignmentVisible` is unused — logic is duplicated instead of reused

`frontend/src/common/assignmentVisibility.js` exports `isAssignmentVisible`, but it isn't imported anywhere. Both `assignments/index.js` and `dashboard/index.js` reimplement the same filter inline:

```js
const releaseDate = assignment.published_date ? moment(assignment.published_date) : null;
return assignment.published && (!releaseDate || !now.isBefore(releaseDate));
```

Suggest calling `isAssignmentVisible(assignment)` in both places instead, which is presumably why the helper was extracted in the first place.

### PR description mentions a change that isn't in the diff

The description lists "Fix JSX structure in CreateAssignment after recent AI settings changes," but `CreateAssignment.js` isn't touched in this diff. Worth confirming that change didn't get dropped, or updating the description.

### Instructor publish indicator is hard to distinguish between "scheduled" and "unpublished"

The button only has two visual states (`primary` for published, `default` otherwise) — "scheduled" vs. "unpublished" is only distinguishable via the native `title` tooltip on hover, which isn't very discoverable. Consider a third visual treatment (color/icon) for "scheduled" so all three states requested in the PR description are actually visible at a glance.

### Minor / nits

- `frontend/src/common/assignmentVisibility.js` is missing a trailing newline.
- A couple of blank lines with trailing whitespace were introduced in `assignments/index.js` — may trip lint rules.
- The `sorter: (a, b) => a.published - b.published` in `instructor/assignments/Assignments.js` still sorts by the raw boolean, so "scheduled" and "published" rows sort identically — consider sorting by `getPublishStatus` instead.
- No unit tests for `isAssignmentVisible`/`getPublishStatus`, despite them encoding date-boundary logic (`null` release date, release-time-exactly-now, etc.) that's easy to get subtly wrong.

SashankN7 (changes requested):
**Medium Issues**
1.  assignments/index.js  duplicates visibility logic instead of using the new utility
You created  assignmentVisibility.js  with  isAssignmentVisible()  for exactly this purpose, but  assignments/index.js  re-implements the filtering logic inline:
// javascript
const visibleAssignments = assignmentsData.filter((assignment) => {
  const releaseDate = assignment.published_date
    ? moment(assignment.published_date)
    : null;
  return assignment.published && (!releaseDate || !now.isBefore(releaseDate));
});
This should use the utility:
// javascript
import { isAssignmentVisible } from "../../common/assignmentVisibility";
const visibleAssignments = assignmentsData.filter(isAssignmentVisible);
Same issue in  dashboard/index.js  — the student filtering logic is also duplicated inline. The whole point of extracting a utility is to avoid this duplication.

2. 50+ commits in the PR — needs squash
The branch has over 50 commits including many unrelated merges from main (PR #312, #334, #339, #340, #341, #342, #346, etc.). This makes the diff extremely hard to review. Please squash the PR-specific commits before merge. The actual PR-specific changes are only 5 files.

3. No tests added for the new visibility utility
 assignmentVisibility.js  contains business logic that determines what students can see — this is the kind of logic that should have unit tests. At minimum:
// javascript
// Tests needed:
// isAssignmentVisible({ published: false }) → false
// isAssignmentVisible({ published: true, published_date: future }) → false
// isAssignmentVisible({ published: true, published_date: past }) → true
// isAssignmentVisible({ published: true, published_date: null }) → true
// getPublishStatus({ published: false }) → "unpublished"
// getPublishStatus({ published: true, published_date: future }) → "scheduled"
// getPublishStatus({ published: true, published_date: past }) → "published"
**Minor Issues**
4. Publish status button has no click handler — confusing UX
In  instructor/assignments/Assignments.js  and  instructor/dashboard/index.js , the publish status is rendered as a  <Button>  with no  onClick :
// jsx
<Button
  type={status === "published" ? "primary" : "default"}
  shape="circle"
  size="small"
  title={status}
>
  {" "}
</Button>
A button that does nothing when clicked is confusing. Either:
- Make it a non-clickable element ( <span>  or  <Tag> )
- Or add a click handler to toggle publish status

5.  getPublishStatus  title tooltip is inconsistent with the button type
When  status === "scheduled" , the button is  type="default"  (gray), but the tooltip says "scheduled". Users might not realize gray means "scheduled" vs "unpublished". Consider using  type="dashed"  for scheduled or adding a different visual indicator.

SashankN7 (changes requested):
**Critical / Should Fix**
1. Missing backend test for student role filtering (security gap)
> The backend now filters assignments by role, but there's no test verifying that students actually only receive published+released assignments. This is the core security behavior of this PR. Add a test case with a student-role enrollment and confirm unpublished/future-scheduled assignments are excluded from the response.
> Similarly, there's no test for the  403 Forbidden  case when a user is not enrolled in the course.
2. Missing  403  test for non-enrolled users
> The endpoint now returns  403  for users not enrolled in the course, but  test_course.py  doesn't cover this case. Add a test where  enrollment_query.filter_by.return_value.first.return_value = None  and assert the response is  403 .

**Important / Should Fix**
3. No  user_id  format validation
> The  user_id  parameter is accepted as-is with no UUID format validation. An attacker could pass arbitrary strings. Consider adding a basic UUID format check:
> // python
> import uuid
> try:
>     uuid.UUID(user_id)
> except ValueError:
>     raise BadRequestError("Invalid user_id format")
4. Inconsistent publish status label rendering
> In  instructor/dashboard/index.js , the Tag uses  publishStatusLabels[status]  (the mapped label). But in  instructor/assignments/Assignments.js , it uses  status.toUpperCase() . These should be consistent — use  publishStatusLabels  everywhere since you already exported it:
> // js
> // Assignments.js — change this:
> {status.toUpperCase()}
> // to this:
> {publishStatusLabels[status]}
5. Frontend  isAssignmentVisible  filter is redundant for students (but not harmful)
> The backend already filters out unpublished/unscheduled assignments for students. The frontend  filter(isAssignmentVisible)  in  assignments/index.js  and  dashboard/index.js  is therefore redundant. This is fine as defense-in-depth, but could confuse future contributors. Consider adding a brief comment explaining it's intentional double-filtering for safety, or remove it and rely solely on the backend.
6.  userInfo?.isStudent  check in  dashboard/index.js  may conflict with backend role logic
> In the student dashboard, you check  userInfo?.isStudent  to decide whether to filter assignments client-side. But the backend already returns only visible assignments for students based on enrollment role. This means the frontend and backend could drift out of sync. If a student's role changes or the backend logic is updated, the dashboard count could become wrong. Consider relying on the backend's filtering consistently.

**Minor / Nice to Have**
7.  assignmentVisibility.test.js  missing trailing newline
> The test file doesn't end with a newline. Minor style issue — most linters flag this.
8.  moment  is deprecated — note for future
>  moment  is used for date comparisons in  assignmentVisibility.js . The project already uses  moment  extensively so this is consistent, but  moment  is in maintenance mode. Not a blocker, but worth noting for a future cleanup (e.g., migrating to  dayjs  or native  Date ).
9.  publishStatusOrder  could use a comment explaining the sort direction
> The object  { unpublished: 0, scheduled: 1, published: 2 }  is used for sorting but the convention (lower = first) isn't documented. A brief comment would help.
10. Test mock setup is complex but correct — just verify it covers the branch
> The  test_get_course_assignments_success  mock is fairly nested ( mock_query.side_effect = [enrollment_query, assignment_query] ). This works but is fragile — if the order of queries in the route changes, the test breaks silently. Consider a comment explaining why two mock queries are needed.

DarioDeSilva (changes requested):
Thanks for working through the last three rounds - the visibility util, the squash, the test coverage, and the status-label consistency are all in good shape now. Went through every prior comment against the current diff; everything from rounds 1-3 is genuinely resolved, not just addressed on paper. One small leftover: `dashboard/index.js` imports `isAssignmentVisible` but never calls it - either use it or drop the import.

One new issue that needs fixing before merge:

**The role check in `get_course_assignments` can be bypassed by passing someone else's `user_id`.** The endpoint trusts whatever `user_id` is in the query string to determine the caller's role - there's no session check tying it to who's actually making the request. A student can call this endpoint with an instructor's or TA's `user_id` (obtainable from `get_course_enrollment`) and get back the unfiltered assignment list, unpublished and scheduled included. That's the exact bug #314 was about, just moved from "no filter at all" to "filter that's trivial to route around."

Also flagging for coordination: this function is also being rewritten in #355, which wires it into the session-based `require_course_role` check instead of a `user_id` param (that's what closes the impersonation gap above - it ties the role check to the actual session rather than a client-supplied value). Since #326 branched before that auth foundation was wired into this route, the two will conflict here and need manual reconciliation, not just a rebase, whichever merges second. Worth syncing on which pattern this endpoint should land on so we're not duplicating auth logic two different ways.

Issue #328, opened 2026-06-26
Extend Student AI Feedback Flow to Use Instructor-Configured Prompts


yczhangcc:
## Description

The student AI feedback flow should be updated so students can request feedback using only the prompts enabled by the instructor for each assignment.

The current system already has a student code editor, AI assistant, run button, version history, auto-save, and submission support. This issue should extend the existing AI assistant behavior instead of rebuilding the student feedback system from scratch.

The main goal is to connect the student-facing AI feedback UI with assignment-level AI settings.

## What Needs To Be Done

- Reuse or extend the existing student AI assistant/chat UI.
- Fetch the enabled AI feedback prompts for the current assignment.
- Display only prompts that are enabled by the instructor.
- Allow the student to select one prompt before requesting AI feedback.
- Send the selected `prompt_id` to the backend.
- Display the returned AI feedback in the existing AI assistant/chat panel.
- Make sure disabled or deleted prompts are not shown to students.
- Make sure students cannot manually control which data is sent to the AI.
- Backend should validate the selected prompt before generating feedback.
- Backend should use instructor-selected input permissions when building the AI request.

## Supported Submission Formats

The AI feedback flow should support multiple student submission formats:

- Code written directly in the in-browser code editor
- Uploaded `.py` file

For code editor submissions, the AI feedback should use the current editor content or saved submission code.

For uploaded `.py` files, the backend should read the uploaded Python file content.

Expected Flow
Student opens the assignment page.
Student opens the AI assistant/chat panel.
Frontend fetches enabled prompts for the assignment.
Student sees only instructor-enabled prompts.
Student selects one prompt.
Frontend sends the selected prompt_id with assignment/submission information.
Backend checks that the prompt is enabled for the assignment.
Backend collects only the input data allowed by instructor settings.
Backend calls the selected AI provider/model.
AI feedback is returned and displayed in the student AI assistant/chat panel.

Acceptance Criteria
Student only sees prompts enabled by the instructor.
Disabled or deleted prompts are not visible to students.
Student can select a prompt and request AI feedback.
Selected prompt_id is sent to the backend correctly.
Backend validates that the selected prompt is enabled.
Backend enforces instructor-selected AI input permissions.
AI feedback works for code editor submissions.
AI feedback works for uploaded .py files.
AI feedback works for uploaded .zip files when source files can be extracted/read.
Backend selects the correct student code source based on submission format.
AI feedback is displayed in the existing AI assistant/chat panel.
[...]

Issue #327, opened 2026-06-26
Add assignment-level AI feedback prompt configuration and input permissions


yczhangcc:
## Description

Extend assignment-level AI feedback settings so instructors can control how AI feedback is configured for each assignment.

This builds on the existing course-level AI provider settings and assignment-level model customization from PR #310/#311. This issue focuses on the next layer: assignment-level feedback prompts, allowed AI input data, and instructor-controlled usage limits.

### Feedback Prompt Settings

Instructors should be able to manage AI feedback prompts per assignment.

Each prompt should include a title/name, instruction text, and enabled/disabled status.

Instructors should be able to add, edit, delete, enable, and disable prompts.

Possible default prompt types include correctness checking, debugging failed tests, edge case review, runtime error explanation, code style review, and algorithmic improvement suggestions.

### AI Input Permissions

Instructors should be able to choose which assignment/submission data may be sent to AI.

Supported input permissions should include assignment description, student solution code, test results, test cases if available, and student output if available.

The backend must enforce these permissions before calling the AI provider. The frontend should not be the only layer controlling what data is sent.

### AI Feedback Usage Controls

Instructors should also be able to configure limits for student AI feedback requests.

The settings should include maximum number of AI feedback requests per student and required thinking time between feedback requests.

Expected behavior:

* Empty/null maximum requests means unlimited requests.
* `0` maximum requests means AI feedback requests are disabled.
* A positive maximum request number limits how many times each student can request feedback.
* `0` thinking time means no wait time.
* A positive thinking time means students must wait that many seconds before asking again.

For this issue, the focus is instructor-side configuration, backend validation, and save/load support. Student-facing countdowns, remaining request display, and disabled button behavior can be handled in follow-up work.

### Backend/API Work

AI feedback-related settings should be kept in a dedicated AI feedback route where possible, instead of mixing all AI logic into assignment or submission routes.

The settings should support assignment-level prompt configuration, allowed input configuration, maximum request limits, and thinking time settings.

## Acceptance Criteria

* Instructor can view and update AI feedback settings for an assignment.
* Instructor can manage feedback prompts per assignment.
* Instructor can select which input data AI may receive.
* Instructor can configure maximum feedback requests per student.
* Instructor can configure thinking time between feedback requests.
* Settings are saved and loaded correctly per assignment.
* Backend validates AI feedback settings before saving.
* Backend enforces selected input permissions when building AI feedback context.
* AI feedback-related endpoints are separated into an AI feedback route where possible.
[...]

PR #329, opened 2026-06-26
Instructor side: add AI feedback prompt and input permission (assignment-level)

yczhangcc:
## Summary

This PR implements Issue #327 by adding assignment-level AI feedback prompt configuration and AI input permission settings.
Instructors can now configure feedback prompts for each assignment and control which assignment/submission data can be included in AI feedback generation.

## Changes
- Added assignment-level AI feedback prompt settings.
- Added AI input permission settings for:
  - assignment description
  - student code
  - test results
  - test cases
  - student output
- Added backend validation and normalization for AI feedback settings.
- Added backend enforcement so only instructor-approved inputs are sent to AI.
- Added endpoints for getting and updating assignment AI settings.
- Updated related AI feedback documentation.

## Important Setup Note

This PR updates the database schema.
Before testing locally, please restart Docker and run the Flask database upgrade:

```bash
docker compose down
docker compose up -d
docker compose exec backend flask db upgrade
```
If your local database is out of date, the new assignment AI feedback fields may not exist yet, and the feature may fail locally. Because apparently databases enjoy needing ceremony before behaving.

Manual Testing
Verified Create Assignment page shows the new AI Feedback Settings section.
Verified default feedback prompts appear when AI feedback is enabled.
Verified prompt title and prompt instruction fields are editable.
Verified AI input permission checkboxes are displayed.
Verified assignment-level AI settings can be saved.
Verified the clean PR no longer includes stress-test files.

Automated Testing  run make test from root folder
Backend
Ran backend tests excluding stress tests:

PYTHONPATH=. python -m pytest test --ignore=test/stress

Result:

213 passed
Frontend

Ran frontend tests:

npm test

Result:

24 test suites passed, 51 tests passed
Notes

This PR focuses on instructor-side assignment AI feedback configuration for Issue #327.

Student-facing prompt selection and chat flow integration are intentionally left for Issue #328.
image
image


SashankN7 (changes requested):
1: Migration chain fix (positive, acknowledge the fix)
> On the migration changes overall:
> Nice catch fixing the revision ID collision —  a1b2c3d4e5f6_add_code_drafts_table.py  and  a1b2c3d4e5f6_add_ai_settings_fields.py  both had the same revision ID on main, which is a real bug. Retargeting code drafts to  92009ca5c92a  and giving AI settings its own  c3d4e5f6a7b8  is the right fix.
> One question: what happens for anyone who already has the old  a1b2c3d4e5f6  AI settings migration stamped in their local DB? Alembic will look for that revision and not find it. The  flask db upgrade  in your setup note should handle it, but it might be worth adding a quick note in the PR body about whether anyone needs to run  alembic stamp head  or if a clean Docker restart (wipe + recreate) is sufficient.
2:  build_feedback_prompt  bug — double  student_code / autograder_results 
>  backend/ai_feedback/integration.py  —  build_feedback_prompt :
> When  feedback_context  is provided, you build  context_text  from  render_feedback_context() , but the prompt still unconditionally includes the old  "Student code:\n{code}\n\nAutograder results:\n{autograder_results}\n\n"  block before  context_text . So if  feedback_context  contains  student_code  and  test_results , the student will see the same data twice in the prompt — once hardcoded, once via the filtered context.
> This partially defeats the purpose of the  allowed_inputs  filtering. The old hardcoded lines should probably be removed or conditionally skipped when  feedback_context  is provided. Something like:
> // python
> if feedback_context is None:
>     context_text = (
>         "Student code:\n"
>         f"{code}\n\n"
>         "Autograder results:\n"
>         f"{autograder_results}\n\n"
>     )
> else:
>     context_text = f"{render_feedback_context(feedback_context)}\n\n"
> And then in the return statement, remove the old  "Student code:\n"  and  "Autograder results:\n"  lines entirely, since they're now handled by  context_text .
3: Duplicate keys in  serialize_assignment_ai_settings 
>  backend/ai_feedback/settings.py  —  serialize_assignment_ai_settings :
> The returned dict has both  feedback_prompts  and  ai_feedback_prompts  pointing to the same object, and  allowed_inputs  /  ai_allowed_inputs  pointing to the same object. Having duplicate keys like this is confusing for consumers — which key should the frontend use? If both are needed for backward compat, add a brief comment explaining that. Otherwise, pick one naming convention (preferably  ai_feedback_prompts  /  ai_allowed_inputs  to match the model columns) and drop the other.
4:  routes/__init__.py  refactor — unnecessary churn
>  backend/routes/__init__.py :
> This refactor changes the import style from  from .user import user  to  from . import user as user_routes  but doesn't change any functionality. This adds merge conflict surface area with any concurrent PRs that touch this file. Unless there's a specific reason for this change, I'd suggest reverting it.
5:  update_assignment  changes from bulk  .update()  to individual  .setattr()  — worth noting the behavior difference
>  backend/routes/assignment.py  —  update_assignment :
> The old code used  .filter_by().update()  which is a bulk SQL UPDATE — it works even if the object isn't loaded into the session. The new code loads the assignment first ( filter_by(...).first() ), then uses  setattr  for each field. This is generally fine but:
> 1. It now issues a SELECT before the UPDATE (minor perf difference, probably negligible).
> 2. More importantly —  setattr  with  hasattr  check won't raise if a caller passes a nonexistent field name (it just silently skips it), whereas the old code would have raised a ProgrammingError from the ORM. That's arguably fine, but worth noting the behavior change.
> Otherwise the refactoring to use  split_ai_settings_payload  +  update_assignment_ai_settings  is clean.
6: Acknowledging the setup note
> On the Important Setup Note:
> Thanks for including the Docker +  flask db upgrade  instructions. This is important since the PR adds two new JSON columns ( ai_feedback_prompts ,  ai_allowed_inputs ) that existing databases won't have. I've verified that the  ADD COLUMN IF NOT EXISTS  in the migration makes it safe to run multiple times.
7: New endpoints lack auth checks
>  backend/routes/assignment.py  —  get_assignment_ai_settings  and  update_assignment_ai_settings_route :
> These new endpoints ( /assignments/<id>/ai-settings ) don't have any auth/authorization checks. Are they only accessed from the instructor-facing UI? If so, that's probably fine for now since the frontend guards it, but it might be worth adding a comment noting that auth is delegated to the frontend and should be hardened when the student-facing side (Issue #328) is implemented.

**In Summary:**

Blockers (requesting changes for these)
1.  build_feedback_prompt  bug — When  feedback_context  is provided, the old hardcoded  "Student code:\n{code}\n\nAutograder results:\n{autograder_results}\n\n"  block is still appended before  context_text . This means  student_code  and  test_results  appear twice in the prompt, partially defeating the  allowed_inputs  filtering. The old hardcoded lines need to be conditionally skipped when  feedback_context  is provided.
2. Duplicate API keys in  serialize_assignment_ai_settings  — The returned dict has both  feedback_prompts  and  ai_feedback_prompts  pointing to the same object, and  allowed_inputs  /  ai_allowed_inputs  pointing to the same object. Consumers won't know which key to use. Pick one naming convention and drop the duplicates.

Some Nice to Haves:
1.  routes/__init__.py  unnecessary refactor — Switching from  from .user import user  to  from . import user as user_routes; app.register_blueprint(user_routes.user)  adds indirection with no benefit. It also creates merge conflict risk with concurrent PRs. Would be cleaner to revert.
2. Migration chain note — The fix for the revision ID collision is correct and good. Just make sure the setup instructions note whether  alembic stamp head  is needed for anyone with an existing DB that already ran the old  a1b2c3d4e5f6  AI settings migration.

SashankN7:
Good work overall, however there are a couple of issues to be addressed:

**Critical Issues**
1. Missing Instructor Authorization on New Routes (Security)
The new  ai_feedback.py  routes have a  # TODO: server-side instructor authorization pending  comment on both  GET  and  PUT  endpoints. This means any authenticated user (including students) could read or modify AI settings on any assignment. This is a privilege escalation vulnerability that must be resolved before merge.
The routes should verify the requesting user is an instructor or TA for the assignment's course, consistent with how other assignment mutation endpoints work.

**Moderate Issues**
2. Inconsistent Import Style in  routes/__init__.py 
The commit changes the existing import pattern from direct blueprint imports:
// python
from .user import user            # existing style
to module-level imports:
// python
from . import user as user_routes  # new style
This is a broader refactor unrelated to the PR's scope. It should either be in a separate commit/PR or at minimum called out. Mixing styles within the same PR makes the diff noisier and harder to review.
3.  ai_feedback_wait_seconds  Allows Zero but Semantics Are Ambiguous
A value of  0  means "no cooldown" but this isn't documented in the UI or the backend validation. Consider either:
- Renaming the label to "Minimum seconds between requests (0 = no limit)"
- Or adding helper text in the  AIFeedbackSettingsSection  component
4. No Upper Bound Validation on  ai_feedback_max_requests 
The  normalize_non_negative_int  helper validates  >= 0 , but there's no reasonable upper bound. A value of  999999  could be set accidentally. Consider adding a sensible max (e.g., 1000) or at least documenting that  null  = unlimited.

**Minor / Style Issues**
5. Migration File Naming
The stat output shows the migration as  b8c9d0e1_add_assignment_ai_feedback_settings.py  but the diff summary referenced  f6a7b8c9d0e1 . Verify the actual filename is consistent and follows the project's migration naming convention.
6. Frontend:  InputNumber  for  ai_feedback_max_requests  Shows  null  as Empty
When the value is  null  (unlimited), the  InputNumber  field will show empty. Consider adding a placeholder like  "No limit"  or a checkbox to toggle between "unlimited" and a numeric value for better UX.
7. Test Assertions Could Be More Specific
The new route tests ( test_ai_feedback_routes.py ) use  assertEqual  on response JSON but some assertions could be more granular — e.g., testing that the response only contains expected keys, not just checking specific field values.

**What's Good** 
- Route separation is clean — moving AI settings out of the monolithic  assignment.py  into a dedicated  ai_feedback.py  blueprint improves maintainability.
- Migration is safe — nullable  max_requests  (unlimited by default) and  wait_seconds  defaulting to 0 won't break existing assignments.
- Validation logic in  normalize_non_negative_int  is solid and well-tested.
- Frontend integration is thorough — all three surfaces (Create, Edit, Assignment Settings) are updated consistently.
- Test coverage is comprehensive — both backend (routes, settings, validation) and frontend (rendering, payload inclusion) are covered.
- Documentation was updated alongside the code.

yczhangcc:
Thanks for the careful review. The comments were helpful, especially the authorization issue on the new AI feedback routes.

I addressed the review comments in the latest push:

Added server-side instructor/TA authorization checks for the AI feedback settings GET/PUT routes.
Added tests for instructor/TA access and student/unrelated-user denial.
Adjusted the route registration style to avoid unnecessary import-pattern changes.
Clarified the wait-time label so 0 clearly means no wait time.
Added a 1000 upper bound for ai_feedback_max_requests.
Added a No limit placeholder for the unlimited max-request case.
Verified the migration filename/revision naming.
Made the AI feedback route response-shape tests more specific.
Updated the README testing notes with separate backend/frontend test commands and the migration upgrade reminder.

I also reran the checks:
python -m pytest test/unit -q: 240 passed
npm test -- --runInBand --watchAll=false: 24 suites passed, 53 tests passed
npm run build: succeeded with existing project-wide warnings
git diff --check: passed

SashankN7:
The critical security issue (missing authorization) is now properly addressed. The code is now clean and well-tested. The import style cleanup in  __init__.py  is a nice improvement. No remaining blockers. Great work, LGTM.

SashankN7 (approved):
lgtm

kiat:
Thank you everyone!

PR #330, opened 2026-06-27
Establish session-based auth as the enforcement boundary

DarioDeSilva:
Splits out the auth-foundation piece of #330 per the 2026-07-13 review request to break this up into smaller PRs. This PR is scoped to just the session/auth mechanics; the TA role backend and frontend follow as separate PRs once this lands.

Note: this branch was rebased against `main` to reconcile with PR #312 (cascade-delete-on-user), which merged separately and independently touched overlapping auth/CORS/SECRET_KEY territory. See the merge commit for the reconciliation details; the summary is in the comments below.

## What changed

- `user_login` and `google_login` write `session["user_id"]` after successful authentication (already covered by #312's merge into `user.py`), establishing a server-side identity that `require_authenticated()`/`require_course_role()` below rely on.
- New `require_authenticated()` and `require_course_role(course_id, allowed_roles, message)` helpers in `util/auth.py`, giving routes one place to check identity and course role. `get_user_course_role()` caches its result per-request on `flask.g` so repeated checks in the same request don't re-query, and treats a `None` role the same as "not enrolled" rather than crashing.
- New `UnauthorizedError` (401) for "not authenticated", separate from `ForbiddenError` (403) for "authenticated but not allowed".
- `SESSION_COOKIE_HTTPONLY`/`SAMESITE`/`SECURE` are now set explicitly (`main` had no session cookie config at all before this), with `SECURE` defaulting to `true` automatically when `SAMESITE=None` is set explicitly.
- Frontend axios instance sends `withCredentials: true` and clears cached auth state + redirects to login on a 401 response.

## What this doesn't do yet

No route uses `require_course_role()` to guard access yet. `course.py`, `assignment.py`, `submission.py`, and `regrade_request.py` are untouched by this PR, so their destructive and admin endpoints have no server-side auth check today, on `main`, and still won't immediately after this merges — this PR neither creates nor closes that exposure. Wiring `require_course_role()` into all of them, which is what actually closes it, is the very next PR, opened as soon as this one merges.

## Testing

Backend: 304 unit tests pass (`pytest test/unit`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,6 +112,7 @@ venv.bak/
 .DS_Store
 backend/.env
 frontend/.env
+backend/.flask_secret_key
 
 # Spyder project settings
 .spyderproject
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,14 @@
+ifeq ($(OS),Windows_NT)
+PYTHON ?= python
+else
+PYTHON ?= python3
+endif
+PIP ?= $(PYTHON) -m pip
+
 .PHONY: install test clean
 install:
 	@echo "Installing backend dependencies..."
-	@pip install -r ./backend/requirements.txt
+	@$(PIP) install -r ./backend/requirements.txt
 	@echo "Backend dependencies installed."
 	@echo "Installing frontend dependencies..."
 	@cd ./frontend && npm install
@@ -11,7 +18,7 @@ test: test-backend test-frontend  ## Run all tests
 
 test-backend:                     ## Backend unit / integration tests
 	@echo "Running backend tests..."
-	@cd ./backend && python3 -m pytest
+	@cd ./backend && $(PYTHON) -m pytest
 	@echo "Backend tests passed."
 
 test-frontend:                    ## Front-end Jest tests
--- a/README.md
+++ b/README.md
@@ -51,6 +51,15 @@ Note: For macOS make sure to turn off Airplay as it uses localport:5000 as well 
     ```
 Note: PASSWORD_SALT is used for password hashing so the same password produces consistent hashed values across local environments. The backend can still run without this value, but it may fall back to less secure password handling depending on the local configuration. For normal development and testing, developers should use the shared project salt in their .env file when available.
 
+    Optional backend environment variables (session auth / CORS):
+
+    ```bash
+    SECRET_KEY="random secret used to sign session cookies; required in production, auto-generated fallback otherwise"
+    FRONTEND_URL="http://localhost:3000"          # comma-separated list of allowed CORS origins
+    SESSION_COOKIE_SAMESITE="Lax"                 # set to "None" if the frontend is on a different domain
+    SESSION_COOKIE_SECURE="false"                 # must be "true" (HTTPS) whenever SameSite is "None"
+    ```
+
 5. run `docker compose up`
 
 6. Visit the pgadmin website via the url in the container. login with the default login:  
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -29,12 +29,18 @@ def _get_or_create_fallback_secret_key():
     across multiple hosts, replicas, or containers with separate
     filesystems and no shared volume; those deployments still need
     SECRET_KEY set explicitly in the environment.
+
+    Returns (key, created_new). created_new is True only when this call
+    generated a fresh key because none was found on disk, which silently
+    invalidates any sessions signed with a previous key (whether this is
+    genuinely the first run, or the persisted file was deleted/corrupted
+    since); callers should log that distinctly from the reuse case.
     """
     try:
         with open(_FALLBACK_SECRET_KEY_PATH, 'r') as f:
             key = f.read().strip()
             if key:
-                return key
+                return key, False
     except FileNotFoundError:
         pass
 
@@ -43,11 +49,11 @@ def _get_or_create_fallback_secret_key():
         fd = os.open(_FALLBACK_SECRET_KEY_PATH, os.O_CREAT | os.O_EXCL | os.O_WRONLY, mode=0o600)
         with os.fdopen(fd, 'w') as f:
             f.write(new_key)
-        return new_key
+        return new_key, True
     except FileExistsError:
         # Another process won the race to create the file; use its value.
         with open(_FALLBACK_SECRET_KEY_PATH, 'r') as f:
-            return f.read().strip()
+            return f.read().strip(), False
 
 
 def create_app(config_class='config.Config'):
@@ -55,13 +61,23 @@ def create_app(config_class='config.Config'):
     app = Flask(__name__)
     app.secret_key = os.getenv('SECRET_KEY')
     if not app.secret_key:
-        app.secret_key = _get_or_create_fallback_secret_key()
-        app.logger.warning(
-            "SECRET_KEY environment variable not set; using a fallback key "
-            "persisted to %s. Set SECRET_KEY in the environment for "
-            "production deployments.",
-            _FALLBACK_SECRET_KEY_PATH,
-        )
+        app.secret_key, created_new_key = _get_or_create_fallback_secret_key()
+        if created_new_key:
+            app.logger.critical(
+                "No fallback SECRET_KEY found at %s; generated a new one. "
+                "Any sessions signed with a previous key (including ones from "
+                "before this process started, if the file was deleted or "
+                "corrupted) are now invalid. Set SECRET_KEY in the environment "
+                "to avoid this.",
+                _FALLBACK_SECRET_KEY_PATH,
+            )
+        else:
+            app.logger.warning(
+                "SECRET_KEY environment variable not set; reusing the fallback "
+                "key persisted to %s. Set SECRET_KEY in the environment for "
+                "production deployments.",
+                _FALLBACK_SECRET_KEY_PATH,
+            )
 
     # Load environment variables
     load_dotenv()
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -1,3 +1,5 @@
+import os
+import secrets
 from flask import Flask
 from flask_marshmallow import Marshmallow
 from flask_sqlalchemy import SQLAlchemy
@@ -10,19 +12,84 @@ ma = Marshmallow()
 db = SQLAlchemy()
 migrate = Migrate()
 
+_FALLBACK_SECRET_KEY_PATH = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__))), '.flask_secret_key'
+)
+
+
+def _get_or_create_fallback_secret_key():
+    """Reads a persisted fallback key, or creates one on first use.
+
+    A pure per-process random key breaks session validation across
+    co-located worker processes (e.g. gunicorn --workers N), since each
+    process would sign cookies with a different key. Persisting it to a
+    local file lets processes on the same host share one.
+
+    This only covers processes sharing one filesystem. It does not help
+    across multiple hosts, replicas, or containers with separate
+    filesystems and no shared volume; those deployments still need
+    SECRET_KEY set explicitly in the environment.
+    """
+    try:
+        with open(_FALLBACK_SECRET_KEY_PATH, 'r') as f:
+            key = f.read().strip()
+            if key:
+                return key
+    except FileNotFoundError:
+        pass
+
+    new_key = secrets.token_hex(32)
+    try:
+        fd = os.open(_FALLBACK_SECRET_KEY_PATH, os.O_CREAT | os.O_EXCL | os.O_WRONLY, mode=0o600)
+        with os.fdopen(fd, 'w') as f:
+            f.write(new_key)
+        return new_key
+    except FileExistsError:
+        # Another process won the race to create the file; use its value.
+        with open(_FALLBACK_SECRET_KEY_PATH, 'r') as f:
+            return f.read().strip()
+
+
 def create_app(config_class='config.Config'):
     # Create the Flask app instance
     app = Flask(__name__)
-    app.secret_key = 'codeassist'
+    app.secret_key = os.getenv('SECRET_KEY')
+    if not app.secret_key:
+        app.secret_key = _get_or_create_fallback_secret_key()
+        app.logger.warning(
+            "SECRET_KEY environment variable not set; using a fallback key "
+            "persisted to %s. Set SECRET_KEY in the environment for "
+            "production deployments.",
+            _FALLBACK_SECRET_KEY_PATH,
+        )
 
     # Load environment variables
     load_dotenv()
-    
+
     # Load the configuration
     app.config.from_object(config_class)
-    
+
+    if not app.config.get('SESSION_COOKIE_SECURE'):
+        app.logger.warning(
+            "SESSION_COOKIE_SECURE is not set to true; session cookies will be "
+            "sent over plain HTTP. Set SESSION_COOKIE_SECURE=true in the "
+            "environment for production and cross-domain deployments."
+        )
+
     # Initialize the extensions with the app
-    CORS(app)
+    # FRONTEND_URL may list multiple comma-separated origins
+    if not os.getenv('FRONTEND_URL'):
+        app.logger.warning(
+            "FRONTEND_URL environment variable not set; CORS is defaulting to "
+            "http://localhost:3000. Set FRONTEND_URL in the environment for "
+            "production deployments."
+        )
+    frontend_origins = [
+        origin.strip()
+        for origin in os.getenv('FRONTEND_URL', 'http://localhost:3000').split(',')
+        if origin.strip()
+    ]
+    CORS(app, supports_credentials=True, origins=frontend_origins)
     ma.init_app(app)
     db.init_app(app)
     migrate.init_app(app, db)
--- a/backend/api/__init__.py
+++ b/backend/api/__init__.py
@@ -1,5 +1,5 @@
 import os
-import secrets
+import logging
 from flask import Flask
 from flask_marshmallow import Marshmallow
 from flask_sqlalchemy import SQLAlchemy
@@ -12,78 +12,23 @@ ma = Marshmallow()
 db = SQLAlchemy()
 migrate = Migrate()
 
-_FALLBACK_SECRET_KEY_PATH = os.path.join(
-    os.path.dirname(os.path.dirname(os.path.abspath(__file__))), '.flask_secret_key'
-)
-
-
-def _get_or_create_fallback_secret_key():
-    """Reads a persisted fallback key, or creates one on first use.
-
-    A pure per-process random key breaks session validation across
-    co-located worker processes (e.g. gunicorn --workers N), since each
-    process would sign cookies with a different key. Persisting it to a
-    local file lets processes on the same host share one.
-
-    This only covers processes sharing one filesystem. It does not help
-    across multiple hosts, replicas, or containers with separate
-    filesystems and no shared volume; those deployments still need
-    SECRET_KEY set explicitly in the environment.
-
-    Returns (key, created_new). created_new is True only when this call
-    generated a fresh key because none was found on disk, which silently
-    invalidates any sessions signed with a previous key (whether this is
-    genuinely the first run, or the persisted file was deleted/corrupted
-    since); callers should log that distinctly from the reuse case.
-    """
-    try:
-        with open(_FALLBACK_SECRET_KEY_PATH, 'r') as f:
-            key = f.read().strip()
-            if key:
-                return key, False
-    except FileNotFoundError:
-        pass
-
-    new_key = secrets.token_hex(32)
-    try:
-        fd = os.open(_FALLBACK_SECRET_KEY_PATH, os.O_CREAT | os.O_EXCL | os.O_WRONLY, mode=0o600)
-        with os.fdopen(fd, 'w') as f:
-            f.write(new_key)
-        return new_key, True
-    except FileExistsError:
-        # Another process won the race to create the file; use its value.
-        with open(_FALLBACK_SECRET_KEY_PATH, 'r') as f:
-            return f.read().strip(), False
-
-
 def create_app(config_class='config.Config'):
     # Create the Flask app instance
     app = Flask(__name__)
-    app.secret_key = os.getenv('SECRET_KEY')
-    if not app.secret_key:
-        app.secret_key, created_new_key = _get_or_create_fallback_secret_key()
-        if created_new_key:
-            app.logger.critical(
-                "No fallback SECRET_KEY found at %s; generated a new one. "
-                "Any sessions signed with a previous key (including ones from "
-                "before this process started, if the file was deleted or "
-                "corrupted) are now invalid. Set SECRET_KEY in the environment "
-                "to avoid this.",
-                _FALLBACK_SECRET_KEY_PATH,
-            )
-        else:
-            app.logger.warning(
-                "SECRET_KEY environment variable not set; reusing the fallback "
-                "key persisted to %s. Set SECRET_KEY in the environment for "
-                "production deployments.",
-                _FALLBACK_SECRET_KEY_PATH,
-            )
 
-    # Load environment variables
+    # Load environment variables first so SECRET_KEY etc. are available
     load_dotenv()
 
-    # Load the configuration
+    # Load the configuration (provides defaults like SECRET_KEY for test config)
     app.config.from_object(config_class)
+
+    # SECRET_KEY must be set via env var or config class for session security.
+    app.secret_key = os.environ.get('SECRET_KEY') or app.config.get('SECRET_KEY')
+    if not app.secret_key:
+        raise RuntimeError(
+            "SECRET_KEY environment variable is required. "
+            "Set it to a random secret string (e.g. python -c 'import secrets; print(secrets.token_hex(32))')."
+        )
 
     if not app.config.get('SESSION_COOKIE_SECURE'):
         app.logger.warning(
@@ -93,19 +38,18 @@ def create_app(config_class='config.Config'):
         )
 
     # Initialize the extensions with the app
-    # FRONTEND_URL may list multiple comma-separated origins
-    if not os.getenv('FRONTEND_URL'):
-        app.logger.warning(
-            "FRONTEND_URL environment variable not set; CORS is defaulting to "
-            "http://localhost:3000. Set FRONTEND_URL in the environment for "
-            "production deployments."
+    # Lock CORS to the actual frontend origin to prevent CSRF on credentialed requests.
+    # FRONTEND_ORIGIN must be set in production; defaults to localhost only for dev.
+    frontend_origin = os.environ.get('FRONTEND_ORIGIN')
+    if not frontend_origin:
+        _logger = logging.getLogger(__name__)
+        _logger.warning(
+            "FRONTEND_ORIGIN environment variable is not set. "
+            "Defaulting to http://localhost:3000 for development only. "
+            "Set FRONTEND_ORIGIN to your actual frontend URL in production."
         )
-    frontend_origins = [
-        origin.strip()
-        for origin in os.getenv('FRONTEND_URL', 'http://localhost:3000').split(',')
-        if origin.strip()
-    ]
-    CORS(app, supports_credentials=True, origins=frontend_origins)
+        frontend_origin = 'http://localhost:3000'
+    CORS(app, supports_credentials=True, origins=[frontend_origin])
     ma.init_app(app)
     db.init_app(app)
     migrate.init_app(app, db)
--- a/backend/api/models.py
+++ b/backend/api/models.py
@@ -19,7 +19,7 @@ class Course(db.Model):
     __tablename__ = "courses"
     id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
     name = db.Column(db.String, nullable=False)
-    instructor_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id"), nullable=False)
+    instructor_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id", ondelete="RESTRICT"), nullable=False)
     sis_course_id = db.Column(db.String, nullable=True)
     semester = db.Column(db.String, nullable=False)
     year = db.Column(db.String, nullable=False)
@@ -42,15 +42,15 @@ class Course(db.Model):
 @dataclass
 class Enrollment(db.Model):
     __tablename__ = "enrollments"
-    student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id"), primary_key=True, nullable=False)
-    course_id = db.Column(UUID(as_uuid=False), db.ForeignKey("courses.id"), primary_key=True, nullable=False)
+    student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id", ondelete="CASCADE"), primary_key=True, nullable=False)
+    course_id = db.Column(UUID(as_uuid=False), db.ForeignKey("courses.id", ondelete="CASCADE"), primary_key=True, nullable=False)
     role = db.Column(db.String, nullable = False)
 
 class Assignment(db.Model):
     __tablename__ = "assignments"
     id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
     name = db.Column(db.String, nullable=False)
-    course_id = db.Column(UUID(as_uuid=False), db.ForeignKey("courses.id"), nullable=False)
+    course_id = db.Column(UUID(as_uuid=False), db.ForeignKey("courses.id", ondelete="CASCADE"), nullable=False)
     due_date = db.Column(TIMESTAMP(timezone=True), nullable=True)
     anonymous_grading = db.Column(db.Boolean, default=False)
     enable_group = db.Column(db.Boolean, default=False)
@@ -92,8 +92,8 @@ class Submission(db.Model):
     file_name = db.Column(db.String, nullable=False)
     submission_number = db.Column(db.Integer, nullable=False)
     submitted_at = db.Column(TIMESTAMP(timezone=True), nullable=True)
-    student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id"), nullable=False, index=True)
-    assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id"), nullable=False, index=True)
+    student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id", ondelete="CASCADE"), nullable=False, index=True)
+    assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id", ondelete="CASCADE"), nullable=False, index=True)
     student_code_file = db.Column(LargeBinary, nullable=False)
     results = db.Column(LargeBinary, nullable=True)
     score = db.Column(db.Float, nullable=True)
@@ -107,13 +107,13 @@ class Submission(db.Model):
 # Handling multiple submitters for a single submission
 class SubmissionSubmitter(db.Model):
     __tablename__ = "submission_submitters"
-    submission_id = db.Column(UUID(as_uuid=False), db.ForeignKey("submissions.id"), primary_key=True, nullable=False)
-    submitter_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id"), primary_key=True, nullable=False)
+    submission_id = db.Column(UUID(as_uuid=False), db.ForeignKey("submissions.id", ondelete="CASCADE"), primary_key=True, nullable=False)
+    submitter_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id", ondelete="CASCADE"), primary_key=True, nullable=False)
     
 class TestCase(db.Model):
     __tablename__ = "test_cases"
     id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
-    assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id"), nullable=False)
+    assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id", ondelete="CASCADE"), nullable=False)
     test_case_name = db.Column(db.String, nullable=False)
     expected_output = db.Column(db.Text, nullable=False)
     input_data = db.Column(db.Text, nullable=False)  
@@ -123,8 +123,8 @@ class TestCase(db.Model):
 class TestCaseResult(db.Model):
     __tablename__ = "test_case_results"
     id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
-    submission_id = db.Column(UUID(as_uuid=False), db.ForeignKey("submissions.id"), nullable=False)
-    test_case_id = db.Column(UUID(as_uuid=False), db.ForeignKey("test_cases.id"), nullable=False)
+    submission_id = db.Column(UUID(as_uuid=False), db.ForeignKey("submissions.id", ondelete="CASCADE"), nullable=False)
+    test_case_id = db.Column(UUID(as_uuid=False), db.ForeignKey("test_cases.id", ondelete="CASCADE"), nullable=False)
     student_output = db.Column(db.Text, nullable=True)
     passed = db.Column(db.Boolean, nullable=True)
 
@@ -134,7 +134,7 @@ class TestCaseResult(db.Model):
 class RegradeRequest(db.Model):
     __tablename__ = "regrade_requests"
     id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
-    submission_id = db.Column(UUID(as_uuid=False), db.ForeignKey("submissions.id"), nullable=False)
+    submission_id = db.Column(UUID(as_uuid=False), db.ForeignKey("submissions.id", ondelete="CASCADE"), nullable=False)
     justification = db.Column(db.Text, nullable=False)
     reviewed = db.Column(db.Boolean, default=False)
 
@@ -143,8 +143,8 @@ class RegradeRequest(db.Model):
 class AssignmentExtension(db.Model):
     __tablename__ = "assignment_extensions"
     id = db.Column(UUID(as_uuid=False), primary_key=True, nullable=False)
-    assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id"), nullable=False)
-    student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id"), nullable=False)
+    assignment_id = db.Column(UUID(as_uuid=False), db.ForeignKey("assignments.id", ondelete="CASCADE"), nullable=False)
+    student_id = db.Column(UUID(as_uuid=False), db.ForeignKey("users.id", ondelete="CASCADE"), nullable=False)
     release_date_extension = db.Column(TIMESTAMP(timezone=True), nullable=True)
     due_date_extension = db.Column(TIMESTAMP(timezone=True), nullable=True)
     late_due_date_extension = db.Column(TIMESTAMP(timezone=True), nullable=True)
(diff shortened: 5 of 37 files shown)


SashankN7 (changes requested):
**Critical Issues**
1. No server-side session/auth check —  requester_id  is trusted from the client (security)
The instructor-only guards check:
// python
requester_id = data.get("requester_id") or request.args.get("requester_id")
if get_user_course_role(requester_id, course_id) != "instructor":
    raise ForbiddenError("Only instructors can perform this action")
But  requester_id  comes directly from the client with no verification that it matches the authenticated user. Any student can send  requester_id: <instructor_uuid>  and bypass all guards. This is a horizontal privilege escalation.
Fix: Use the session-based auth that already exists (the  session["user_id"]  check from PR #312's  _verify_student ). Every guarded endpoint should verify:
// python
session_user = session.get("user_id")
if not session_user:
    raise ForbiddenError("Not authenticated")
if session_user != requester_id:
    raise ForbiddenError("requester_id does not match session user")
Or better: derive  requester_id  from the session and ignore the client-provided value entirely.
2. Frontend-only guards are bypassable (defense-in-depth concern)
The sidebar gating, redirect guards in courseSettings/aiSettings, and hidden buttons are all client-side. A student can directly navigate to  /courseSettings  or call the API with  requester_id  set to an instructor UUID. The backend guards (issue #1) are the only real security boundary — and they have the  requester_id  trust issue above.
**Significant Issues**
3.  get_user_enrollments  returns  enrollment_role  for ALL enrollments, not just the caller's
The endpoint returns every enrollment for a given  user_id :
// python
enrollments = db.session.query(Enrollment).filter_by(student_id=user_id)
Each enrollment now includes  enrollment_role . This is fine for the caller's own data, but if  user_id  is ever provided by another user (e.g., an admin viewing enrollments), it leaks all role information. Not critical but worth noting.
4.  get_my_enrollment_role  doesn't verify the caller is enrolled
The endpoint:
// python
@course.route("/get_my_enrollment_role", methods=["GET"])
def get_my_enrollment_role():
    user_id = request.args.get("user_id")
    course_id = request.args.get("course_id")
    ...
    enrollment = db.session.query(Enrollment).filter_by(student_id=user_id, course_id=course_id).first()
Returns 404 "Enrollment not found" for unenrolled users — which leaks whether a specific user is enrolled in a specific course. Consider returning a generic "not authorized" or empty role instead.
5.  sider.js  uses raw  fetch  — potential inconsistency
The raw  fetch  to  /get_my_enrollment_role  bypasses the axios instance's base URL configuration, auth interceptors, and error handling. If the base URL ever changes (e.g., different API host), this fetch will break silently. Consider adding the API base URL from config.
6. Migration change to  op.execute  with raw SQL
The migration switches from  op.add_column  to:
// python
op.execute("ALTER TABLE assignments ADD COLUMN IF NOT EXISTS enable_code_editor BOOLEAN DEFAULT FALSE")
This is PostgreSQL-specific (not portable to SQLite/MySQL). If the project ever needs cross-DB support, this will break. The original  op.add_column  was more portable. Minor concern if PostgreSQL is the only target.
7. No test for the negative path — non-instructor calling guarded endpoints
The tests mock  get_user_course_role  to return  "instructor"  for all guarded calls, but there's no test verifying that a non-instructor is rejected (403 Forbidden). Add tests for:
- Student calling  delete_assignment  → 403
- TA calling  update_ai_settings  → 403
- Unauthenticated call → 403
🟢 Minor / Style Issues
8.  RolePill  component is defined inline in enrollment/index.js
Extract it to a shared component file if it's used elsewhere, or keep it inline if only used in enrollment. Minor, but the enrollment file is already large.
9.  accountPopoverContent.js  clears  courseRole  from localStorage
Good — but should also clear it from  GlobalContext . Verify  updateCourseRole(null)  or equivalent is called on logout.
10. Co-authored-by credit
Commit  c2a50eb  has  Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>  — fine for attribution, just noting it.

SashankN7:
**Critical Issues**
1. No server-side session/auth check —  requester_id  is trusted from the client (security)
The instructor-only guards check:
// python
requester_id = data.get("requester_id") or request.args.get("requester_id")
if get_user_course_role(requester_id, course_id) != "instructor":
    raise ForbiddenError("Only instructors can perform this action")
But  requester_id  comes directly from the client with no verification that it matches the authenticated user. Any student can send  requester_id: <instructor_uuid>  and bypass all guards. This is a horizontal privilege escalation.
Fix: Use the session-based auth that already exists (the  session["user_id"]  check from PR #312's  _verify_student ). Every guarded endpoint should verify:
// python
session_user = session.get("user_id")
if not session_user:
    raise ForbiddenError("Not authenticated")
if session_user != requester_id:
    raise ForbiddenError("requester_id does not match session user")
Or better: derive  requester_id  from the session and ignore the client-provided value entirely.
2. Frontend-only guards are bypassable (defense-in-depth concern)
The sidebar gating, redirect guards in courseSettings/aiSettings, and hidden buttons are all client-side. A student can directly navigate to  /courseSettings  or call the API with  requester_id  set to an instructor UUID. The backend guards (issue #1) are the only real security boundary — and they have the  requester_id  trust issue above.
**Significant Issues**
3.  get_user_enrollments  returns  enrollment_role  for ALL enrollments, not just the caller's
The endpoint returns every enrollment for a given  user_id :
// python
enrollments = db.session.query(Enrollment).filter_by(student_id=user_id)
Each enrollment now includes  enrollment_role . This is fine for the caller's own data, but if  user_id  is ever provided by another user (e.g., an admin viewing enrollments), it leaks all role information. Not critical but worth noting.
4.  get_my_enrollment_role  doesn't verify the caller is enrolled
The endpoint:
// python
@course.route("/get_my_enrollment_role", methods=["GET"])
def get_my_enrollment_role():
    user_id = request.args.get("user_id")
    course_id = request.args.get("course_id")
    ...
    enrollment = db.session.query(Enrollment).filter_by(student_id=user_id, course_id=course_id).first()
Returns 404 "Enrollment not found" for unenrolled users — which leaks whether a specific user is enrolled in a specific course. Consider returning a generic "not authorized" or empty role instead.
5.  sider.js  uses raw  fetch  — potential inconsistency
The raw  fetch  to  /get_my_enrollment_role  bypasses the axios instance's base URL configuration, auth interceptors, and error handling. If the base URL ever changes (e.g., different API host), this fetch will break silently. Consider adding the API base URL from config.
6. Migration change to  op.execute  with raw SQL
The migration switches from  op.add_column  to:
// python
op.execute("ALTER TABLE assignments ADD COLUMN IF NOT EXISTS enable_code_editor BOOLEAN DEFAULT FALSE")
This is PostgreSQL-specific (not portable to SQLite/MySQL). If the project ever needs cross-DB support, this will break. The original  op.add_column  was more portable. Minor concern if PostgreSQL is the only target.
7. No test for the negative path — non-instructor calling guarded endpoints
The tests mock  get_user_course_role  to return  "instructor"  for all guarded calls, but there's no test verifying that a non-instructor is rejected (403 Forbidden). Add tests for:
- Student calling  delete_assignment  → 403
- TA calling  update_ai_settings  → 403
- Unauthenticated call → 403
**Minor / Style Issues**
8.  RolePill  component is defined inline in enrollment/index.js
Extract it to a shared component file if it's used elsewhere, or keep it inline if only used in enrollment. Minor, but the enrollment file is already large.
9.  accountPopoverContent.js  clears  courseRole  from localStorage
Good — but should also clear it from  GlobalContext . Verify  updateCourseRole(null)  or equivalent is called on logout.
10. Co-authored-by credit
Commit  c2a50eb  has  Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>  — fine for attribution, just noting it.

DarioDeSilva:
Addressed all the review feedback, ready for another look when you get a chance.

yczhangcc (commented):
HI,the original critical requester_id spoofing issue looks fixed for the guarded destructive/admin endpoints. 
A few things I noticed while testing:

1. /get_my_enrollment_role still takes user_id from the query string and does not compare it to session["user_id"]. The destructive/admin guards are fixed, but this endpoint can still expose another user’s course role/enrollment state.
2. /get_user_enrollments still accepts an arbitrary user_id and now returns enrollment_role, so the role/enrollment privacy concern remains for this endpoint as well.
3. In frontend/src/components/layout/sider.js, role hydration uses process.env.REACT_APP_API_URL directly and fails silently if that environment variable is unset. Since courseRole is persisted in localStorage, stale role UI is possible after refresh or direct navigation.

SashankN7 (changes requested):
**Comment 1 — Security:  get_my_enrollment_role  leaks any user's role**
 backend/routes/course.py  —  get_my_enrollment_role 
**Security — still accepting `user_id` from the client**
 
This endpoint takes `user_id` from the query string, meaning any authenticated user can query any other user's role in any course:
 
GET /get_my_enrollment_role?user_id=<victim-uuid>&course_id=<course-uuid>
 
This should derive `user_id` from the session, like the other guards in this PR:
 
```python
def get_my_enrollment_role():
    user_id = session.get("user_id")
    if not user_id:
        raise ForbiddenError("Not authenticated")
    course_id = request.args.get("course_id")
    ...
Same issue applies to  get_user_enrollments  — it also accepts an arbitrary  user_id  and returns that user's full enrollment list with roles. Both endpoints should use  session.get("user_id")  instead of  request.args.get("user_id") .
 
---
 

**Comment 2 — Bug: `delete_submissions` skips auth when assignment not found**
 
**File:** `backend/routes/assignment.py` — `delete_submissions`
 
Bug — auth check is skipped when assignment doesn't exist
If the assignment doesn't exist,  assignment_for_auth  is  None , the  if assignment_for_auth:  block is skipped, and the function continues to delete submissions without any authorization check:
// python
assignment_for_auth = db.session.query(Assignment).filter_by(id=assignment_id).first()
if assignment_for_auth:                          # ← skipped when None
    enrollment = db.session.query(Enrollment)...
    if not enrollment or enrollment.role != "instructor":
        raise ForbiddenError(...)
 
# continues to delete submissions without auth  ← problem
Fix: fail early if the assignment isn't found, before attempting the delete:
// python
assignment_for_auth = db.session.query(Assignment).filter_by(id=assignment_id).first()
if not assignment_for_auth:
    raise NotFoundError("Assignment not found")
 
enrollment = db.session.query(Enrollment).filter_by(
    student_id=requester_id, course_id=assignment_for_auth.course_id
).first()
if not enrollment or enrollment.role != "instructor":
    raise ForbiddenError("Only instructors can delete all submissions")
 
---

 
**Comment 3 — Code quality: `sider.js` uses raw `fetch()` instead of `service()`**
 
**File:** `frontend/src/components/layout/sider.js`
 
Consistency — raw  fetch()  bypasses the  service()  helper
This PR adds a raw  fetch()  call here for  get_my_enrollment_role :
// js
fetch(`${process.env.REACT_APP_API_URL}/get_my_enrollment_role?...`)
The rest of the codebase uses the  service()  helper (from  services/index.js ) which handles auth interceptors, error normalization, and base URL configuration. Using raw  fetch()  here means:
1. If the user's session expires, this call silently fails (no interceptor redirect)
2. Error handling is inconsistent —  .catch(() => {})  swallows everything
Use the service helper instead:
// js
import { getMyEnrollmentRole } from "../../services/course";
// ...
const res = await getMyEnrollmentRole({ user_id: userInfo.id, course_id: courseInfo.id });
if (res?.data?.role) updateCourseRole(res.data.role);
Note: once  get_my_enrollment_role  is fixed to derive  user_id  from the session (Comment 1), you won't even need to pass  user_id  here.
 
---
 
**Comment 4 — Dead code: frontend still sends `requester_id` to backend**

 
**File:** `frontend/src/services/assignment.js`, `frontend/src/pages/instructor/aiSettings/index.js`, `frontend/src/pages/instructor/courseSettings/index.js`, `frontend/src/pages/instructor/enrollment/index.js`
 
Dead code —  requester_id  is sent from the frontend but ignored by the backend
Several frontend files still send  requester_id  as a parameter:
-  deleteAssignment(assignmentId, requesterId)  →  assignment.js 
-  deleteSubmissions(assignmentId, requesterId)  →  assignment.js 
-  deleteAllAssignments({ ..., requester_id: userInfo.id })  →  courseSettings/index.js 
-  deleteCourse({ ..., requester_id: userInfo.id })  →  courseSettings/index.js 
-  updateRole({ ..., requester_id: userInfo.id })  →  enrollment/index.js 
-  updateAiSettings({ ..., requester_id: userInfo.id })  →  aiSettings/index.js 
But the backend now derives  requester_id  from  session.get("user_id")  — the query/body  requester_id  is never read. This is confusing for future maintainers and could give a false sense that client-supplied  requester_id  matters.
Clean up: remove the  requester_id  parameter from all frontend service calls and function signatures, or have the backend read it from session and ignore the client value (with a comment explaining why).
 
---
 
**Comment 5 — Security: migration uses raw SQL instead of Alembic API**

 
**File:** `backend/migrations/versions/b2c3d4e5f6a7_add_enable_code_editor.py`
 
Migration portability — raw SQL breaks SQLite (used in tests)
The migration was changed from the Alembic API to raw SQL:
// python
# Before (portable):
op.add_column('assignments', sa.Column('enable_code_editor', sa.Boolean(), server_default='false', nullable=False))
 
# After (Postgres-only):
op.execute("ALTER TABLE assignments ADD COLUMN IF NOT EXISTS enable_code_editor BOOLEAN DEFAULT 'false' NOT NULL")
 ADD COLUMN IF NOT EXISTS  is Postgres-specific syntax. If the test suite ever uses SQLite (which many Flask test setups do), this migration will fail.
Recommendation: use the Alembic  op.add_column()  API, which generates portable SQL for the target database. If the concern was idempotency on re-run, handle that at the application level rather than in the migration.
 
---
 
**Comment 6 — Minor: `accountPopoverContent.js` doesn't clear `courseRole` from context on logout**

 
**File:** `frontend/src/components/layout/accountPopoverContent.js`
 
Minor —  courseRole  is cleared from localStorage but not from React context
The logout handler clears  localStorage.removeItem("courseRole")  and calls  updateCourseRole("") , which is good. But if a user logs out and logs in as a different user in the same browser session, the stale  courseRole  from GlobalContext might briefly persist until the next course navigation triggers a re-fetch in  sider.js .
This is low-risk since  sider.js  does re-fetch on course mount, but worth noting for completeness — the pattern should be consistent with how  updateUserInfo(null)  handles state cleanup.
 
---
 
**Comment 7 — Missing test: `get_my_enrollment_role` when user is not enrolled**
 

Missing test —  get_my_enrollment_role  for non-enrolled user
The PR adds negative-path tests for  delete_assignment  and  delete_submissions  (unauthenticated, student, TA), but there's no test for  get_my_enrollment_role  when:
1. The requesting user is not enrolled in the course
2. The requesting user is enrolled but the target  user_id  (once fixed to use session) doesn't match
Suggest adding:
// python
def test_get_my_enrollment_role_not_enrolled(client, mocker):
    mock_query = mocker.patch("routes.course.db.session.query")
    mock_query.return_value.filter_by.return_value.first.return_value = None
 
    with client.session_transaction() as sess:
        sess["user_id"] = "student-uuid"
 
    resp = client.get("/get_my_enrollment_role", query_string={"course_id": "course-123"})
    assert resp.status_code == 403
 
---
 
**Comment 8 — TA cannot promote students to TA (by design, but undocumented)**
 

Observation — TA role restrictions may need documentation
The auth guard  get_user_course_role(requester_id, ...) != "instructor"  means TAs cannot:
- Delete assignments or submissions
- Update course settings
- Change enrollment roles
- Update AI settings
This is likely intentional, but the PR description doesn't explicitly state what TAs can vs can't do. Consider adding a brief section to the PR description or a comment in the code listing the TA permission boundaries, so future reviewers don't have to reverse-engineer them from the  != "instructor"  checks.

SashankN7 (changes requested):
**Security**
1. Secret key default is insecure
> The fallback  secret_key = 'codeassist'  in  backend/api/__init__.py  means any deployment that forgets to set  SECRET_KEY  in the environment will use a publicly-known secret. This allows session forgery. Consider either:
> - Generating a random key at startup and failing loudly if not configured, or
> - Using a randomly-generated default at startup (e.g.,  os.urandom(24).hex() ) so at least each deployment gets a unique key — though this breaks across restarts.
> At minimum, add a warning log when the fallback is used.
2.  rerun_submission_autograder  has no authorization check
> The existing  upload_submission  and  get_results  routes also lack authorization, so this is consistent — but since you're actively adding auth guards elsewhere in this PR (great!), I'd flag that  rerun_submission_autograder  is also exploitable: any logged-in user (or even unauthenticated, since it doesn't check  session ) can rerun the autograder for any submission by guessing/iterating  submission_id . Consider adding at minimum a session check and ideally verifying the requester owns the submission.
3.  infer_api_key_provider  pattern matching is fragile
> The prefix-based detection ( sk-ant-  → Claude,  sk-  → OpenAI,  AIza  → Gemini) in  integration.py  will break if providers change their key formats. This is okay as a heuristic, but the  validate_assignment_provider_credential  function logs a warning and proceeds anyway when it can't infer the provider — meaning a Gemini key could silently be sent to OpenAI. Consider at least raising an error when inference fails rather than letting it through.
4.  update_role  doesn't validate the new role value
> The endpoint lowercases  data["new_role"]  but doesn't check it's a valid role ( student ,  ta ,  instructor ). A user could set any arbitrary string. Add validation:
> // python
> valid_roles = {"student", "ta", "instructor"}
> if data["new_role"].lower() not in valid_roles:
>     raise BadRequestError("Invalid role")
5.  update_role  doesn't prevent self-demotion or promoting others above your role
> An instructor can demote themselves or promote someone else to instructor, which could be used to lock out the real instructor. Consider:
> - Preventing self-demotion (or at least requiring confirmation)
> - Preventing TAs from changing any roles (already done ✓) and instructors from promoting TAs to instructor
**Bugs**
6. TA can see the instructor's "Role Management" dropdown on the enrollment page
> In  enrollment/index.js , the TA check prevents TAs from changing roles but the dropdown still renders for TA users viewing other TA or student records. The condition at lines ~233–249 should also hide the dropdown entirely when  isTA  is true, not just disable the "instructor" option. Right now TAs can potentially click the dropdown and attempt role changes (which would fail server-side, but the UI shouldn't allow it).
7.  get_user_enrollments  doesn't filter out deleted/archived courses
> The endpoint joins against all  Course  records matching the enrollment IDs. If a course is soft-deleted or archived, it will still appear in the enrollment list. Verify this is intentional.
8.  rerun_submission_autograder  doesn't clean up Docker containers on all failure paths
> If the code fails between container creation (line 550) and the timeout/error handling blocks, the container could be orphaned. The  finally  block or a  try/finally  wrapper around the container lifecycle would prevent orphaned containers. This is especially important under high load.
**Suggestions**
9. Extract  get_user_course_role  to a shared utility
>  get_user_course_role  is defined in  course.py  but is needed across  assignment.py  too (where it would need to be re-implemented or imported). Since multiple routes use it for authorization, consider moving it to a shared location like  util/auth.py  or  api/auth.py .
10. The  courseRole  localStorage fallback should clear on logout
>  accountPopoverContent.js  clears  courseRole  from context and localStorage on logout ✓ — but if a user's session expires server-side while they're still browsing, their stale  courseRole  from localStorage will persist. The  sider.js  fetch handles re-hydration, but consider also catching 401/403 responses and clearing the stored role.
11.  post_gemini_with_retry  should respect the  assignment  for its credential lookup
> The retry logic for Gemini is well-implemented with exponential backoff. However, since this PR adds per-assignment API keys, confirm that the retry path also re-reads the assignment credential (not just a cached value) in case the key was rotated between attempts.
12.  infer_api_key_provider  should be its own util module
> This function is useful and could be called from assignment creation, AI settings, and the code editor. Consider moving it to  util/  or  ai_feedback/  to avoid duplication.
13. Default model change from  gpt-4o-mini  to  gpt-4o  in AI Settings
> This is a deliberate change but worth calling out —  gpt-4o  is significantly more expensive than  gpt-4o-mini . If this is the default for all new courses, instructors who don't change it will incur higher costs. Verify this is intentional.
14. Consider adding TA role tests
> The PR description lists manual testing steps but no automated tests for TA permissions. Since the role system affects routing, sidebar rendering, and enrollment management, adding at least a few unit/integration tests would prevent regressions. For example:
> - TA can access enrollment page but cannot change roles
> - TA is redirected away from course settings
> - TA sees instructor dashboard navigation
15. The migration revision ID collision fix should be documented
> The merge migration ( 1bdb41066778 ) fixes a revision ID collision. This is good, but consider adding a comment in the migration explaining why the collision occurred to prevent it from happening again with future migrations.
**Things that look good**
- The  courseRole  architecture (per-course, localStorage-persisted, hydrated on entry) is well-designed and handles course-switching correctly.
- The CORS cleanup (removing  @cross_origin()  per-route in favor of global config) is the right approach.
- Session security settings ( HttpOnly ,  SameSite=Lax ) are good additions.
- The  withCredentials: true  on the axios instance is necessary and correctly implemented.
- The  rerun_submission_autograder  feature is well-structured with proper Docker lifecycle management (mostly — see item 8).
- Role pill styling with  RolePill  component is a nice UX improvement over raw  <Select> .

yczhangcc:
please remove frontend/package-lock.json before merge , thanks

DarioDeSilva:
Removed frontend/package-lock.json from the PR in 0a353f0. It was already in .gitignore but had been force-added earlier; it is now untracked.

DarioDeSilva:
Since my last update the branch has moved off client-supplied requester_id entirely and every row of the permission table is now enforced on the backend. Point by point against the review:

1. **requester_id trust (critical):** Fixed. user_login and google_login now establish a Flask session and every guarded endpoint derives the caller from session["user_id"]. The client-supplied requester_id is gone from the routes and the frontend services. There is a regression test confirming that a student passing an instructor's requester_id still gets 403.

2. **Frontend-only guards:** The backend is now the real boundary. Instructor-only: delete_course, delete_assignment, delete_submission, delete_submissions, delete_all_assignments, update_course, update_role, update_ai_settings. Staff (instructor or TA): create/update/duplicate_assignment, upload_assignment_autograder, create/delete_extension, update_grade, set_reviewed, get_instructor_regrade_requests, get_course_enrollment, create_enrollment, create_enrollment_csv. TAs adding users are capped to the student role, including per-row in CSV imports.

3. **get_user_enrollments leaking roles:** It no longer accepts a user_id parameter. It returns only the session user's own enrollments.

4. **get_my_enrollment_role enumeration:** Now session-derived, and unenrolled callers get a generic 403 "Not authorized" instead of the 404 that leaked enrollment status.

5. **sider.js raw fetch:** It builds its URL from REACT_APP_API_URL, the same source the axios instance uses, and sends credentials. A failed lookup now resets courseRole instead of leaving a stale value from a previously visited course. Relatedly, the axios instance sends credentials app-wide, which is why the per-route @cross_origin() decorators were replaced with a single credentialed CORS config.

6. **Postgres-only migration SQL:** Replaced with a SQLAlchemy inspector check plus op.add_column, so it stays idempotent and is no longer tied to Postgres.

7. **Missing negative-path tests:** Added across the board. Unauthenticated calls now return 401 (new UnauthorizedError), authenticated-but-wrong-role calls return 403, and there are TA-allowed positives for grading and assignment creation. Backend suite is at 354 tests, frontend at 58, all passing.

8. **RolePill inline:** Extracted to components/RolePill.js.

Two deployment notes: unauthenticated requests now return 401 and the frontend clears cached auth state and redirects to login when it sees one, so anyone logged in before this change gets logged out once. There are also new optional env vars (SECRET_KEY, comma-separated FRONTEND_URL origins, SESSION_COOKIE_SAMESITE, SESSION_COOKIE_SECURE), documented in the README for cross-domain deployments.

SashankN7 (changes requested):
**Blocking Issues**
1. Fallback  SECRET_KEY  file has a race condition and no permission control
In  backend/api/__init__.py , the  _get_or_create_fallback_secret_key()  function reads/writes a file to disk. Two issues:
- Race condition: If two processes start simultaneously and the file doesn't exist yet, both could generate different keys. Use file locking ( fcntl.flock  on Linux, or  atomicfile  pattern) or generate once and require it exists before startup.
- File permissions: The generated  .flask_secret_key  file is created with default umask permissions (often  644 ), meaning any user on the system can read it. Use  os.open()  with  0o600  mode to restrict permissions:
// python
fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
with os.fdopen(fd, 'w') as f:
    f.write(key)
2.  get_user_course_role  queries the DB but doesn't handle the case where the course doesn't exist
In  backend/util/auth.py ,  get_user_course_role(user_id, course_id)  returns  None  if no enrollment is found. But if the  course_id  itself doesn't exist in the database, a caller could get a confusing "not enrolled" error instead of "course not found." Consider whether  require_course_role  should first verify the course exists, or at minimum document this behavior.
3.  SESSION_COOKIE_SECURE  defaults to  false  — dangerous in production
In  backend/config.py , the line  os.getenv('SESSION_COOKIE_SECURE', 'false').lower() == 'true'  means sessions are sent over HTTP by default. If a production deployment forgets to set this env var, session cookies are vulnerable to MITM attacks. Consider:
- Defaulting to  true  when  SESSION_COOKIE_SAMESITE  is  None  (as the comment already notes)
- Or at minimum, logging a loud warning at startup if both  Secure  is false and  SameSite  is  None 
**Suggestions (Non-blocking)**
4.  _login_as  helper is duplicated across test files
The  _login_as(client, user_id)  function appears in multiple test files ( test_assignment.py ,  test_ai_feedback_routes.py ,  test_course.py , etc.). Extract it into a shared conftest fixture:
// python
# backend/test/unit/conftest.py
@pytest.fixture
def login_as(app, client):
    def _login(user_id):
        with client.session_transaction() as sess:
            sess['user_id'] = user_id
    return _login
This reduces duplication and ensures consistent auth mocking.
5.  require_course_role  runs a DB query on every call — consider caching for the request lifecycle
 require_course_role  calls  get_user_course_role  which hits the database. In routes where multiple auth checks might happen (e.g.,  upload_submission  which checks enrollment separately), this results in redundant queries. Consider caching the role in the Flask  g  object for the duration of the request:
// python
def get_user_course_role(user_id, course_id):
    if hasattr(g, '_course_role_cache') and course_id in g._course_role_cache:
        return g._course_role_cache[course_id]
    # ... query ...
    g._course_role_cache = getattr(g, '_course_role_cache', {})
    g._course_role_cache[course_id] = role
    return role
6.  update_role  prevents self-demotion but doesn't prevent self-promotion
The  update_role  route in  course.py  blocks instructors from demoting themselves, but doesn't block a TA from promoting themselves to instructor via the API. The frontend hides the UI, but the API endpoint should also validate that only instructors can grant instructor or TA roles:
// python
if new_role in ('instructor', 'ta') and requester_role != 'instructor':
    raise ForbiddenError("Only instructors can assign instructor or TA roles")
7. Frontend  courseRole  state is lost on page refresh
The  courseRole  is stored in  localStorage  but only set when navigating to a course via the dashboard. If a user navigates directly to a course URL (e.g., bookmarked),  courseRole  won't be populated. The  sider.js  or  App.js  should fetch the role on mount if  courseInfo  exists but  courseRole  doesn't.
8. CORS  FRONTEND_URL  defaults to  localhost:3000  — silent security downgrade
In  backend/api/__init__.py , if  FRONTEND_URL  isn't set, CORS defaults to  http://localhost:3000 . In production, this means any request from localhost is allowed with credentials. Consider requiring  FRONTEND_URL  in production (similar to how  SECRET_KEY  is now required).
**Positives**
- Session-based auth replacing  requester_id  is a significant security improvement
- The  require_authenticated  /  require_course_role  pattern is clean and reusable
- Good test coverage for negative paths (unauthorized, wrong role, etc.)
- The idempotent migration fix is a nice catch
-  RolePill  component is well-designed for reuse

DarioDeSilva:
Follow-up on the 2026-07-09 review:

- SESSION_COOKIE_SECURE default: added a startup warning when it resolves to false, same pattern as the existing SECRET_KEY fallback warning.
- FRONTEND_URL CORS default: added a startup warning when it falls back to localhost:3000.
- Per-request role caching: get_user_course_role now caches its result on Flask's g object per (user_id, course_id), so repeated auth checks in the same request don't re-query.
- Test helper duplication: added backend/test/unit/conftest.py with a shared login_as fixture, removed the duplicated _login_as definitions from test_assignment.py, test_ai_feedback_routes.py, and test_regrade_request.py, and converted test_course.py's inline session_transaction blocks to the same fixture.
- Added a negative-path test for update_role: a student calling it now has explicit 403 coverage alongside the existing TA case.

Deliberately not changing get_user_course_role/require_course_role to distinguish "course not found" from "not enrolled". Both cases already return a generic 403 today, so there's no information leak. Adding the distinction requires an extra DB query on every auth check, and this test suite mocks db.session.query as a single shared mock with ordered side_effect lists per test, so inserting that query shifts every later call in ~20 tests across test_course.py. Rewriting all of those for an error-message clarity improvement (not a security fix) isn't worth the diff size. Leaving this as-is; happy to revisit if a real leak scenario comes up.

SashankN7 (changes requested):
**1. PR scope is too large — consider splitting**
> This PR touches 117 files with 18k+ insertions and 21k+ deletions. It mixes several unrelated concerns: TA role enforcement, auth security hardening, frontend TA UI, test infrastructure, CORS changes, and more. Consider splitting into smaller PRs (e.g., one for auth/security fixes, one for TA role backend, one for TA role frontend) so each can be reviewed more thoroughly and merged independently.
**2. Migration uses non-portable raw SQL**
> Several migrations use  ALTER TABLE ... ADD COLUMN IF NOT EXISTS  which is PostgreSQL-specific and not portable across databases. If you ever need to support SQLite (e.g., for local dev or testing), these will break. Consider using Alembic's  batch_alter_table  context manager instead.
**3. In-memory role caching could cause stale data**
> The PR mentions caching role lookups for performance, but in-memory caching (e.g., a dict in the module scope) can cause stale data after a role update via  update_role . If an instructor promotes a student to TA, the cached role won't reflect that until the server restarts. Consider either using a short TTL cache, invalidating on write, or just hitting the DB (the query is simple enough).
**4.  get_my_enrollment_role  still takes  user_id  from query string**
> Even though session-based auth was added,  get_my_enrollment_role  still accepts  user_id  as a query parameter. A logged-in user could pass any UUID and learn someone else's role in a course. This should derive  user_id  from  session["user_id"]  exclusively and reject any client-provided  user_id .
**5. Frontend relies on  localStorage  for role — fragile**
> The frontend sider component reads the user's role from  localStorage , which can become stale after a role change or re-enrollment. Consider fetching the current role from the server on app load or using the existing  get_my_enrollment_role  endpoint.

Nice work overall; none of these are blockers, but it'd be better to address them before merging, so requesting changes here.

DarioDeSilva:
Splitting this PR into three per the review feedback above — it had grown from the TA role feature into three separable things: session-based auth, TA role backend enforcement, and TA role frontend UI.

This PR (#330) is now scoped to just the auth foundation: session wiring, the require_authenticated/require_course_role helpers, CORS, and the SECRET_KEY/cookie hardening. The TA role backend and frontend will follow as separate PRs once this one merges, in that order, since both depend on the helpers introduced here.

Branch history before the split is preserved at `backup/pr-330-full-2026-07-13` if useful for reference.

On today's review:

1. PR scope — addressed by this split.
2. Migration raw SQL — the enable_code_editor migration touched by this PR (and its follow-up) already moved off raw SQL onto an inspector check + op.add_column, in an earlier round. The other `ADD COLUMN IF NOT EXISTS` migrations are pre-existing on main and untouched by this PR.
3. In-memory role caching — it's flask.g-scoped (per-request), not a module-level dict, so there's no stale-after-restart concern.
4. get_my_enrollment_role / get_user_enrollments still taking user_id from the query string — both already derive user_id from the session only; that was fixed in an earlier round. Let me know if you're seeing a stale diff view.
5. Frontend localStorage staleness — agreed, worth tightening. That logic lives in the TA-role-frontend PR, so I'll pick it up there.

SashankN7:
**Blocking / Security Issues**
1.  require_course_role  is never called in any route
The new  require_authenticated()  and  require_course_role()  helpers in  util/auth.py  are defined and tested, but not integrated into any existing route. The PR title says "Establish session-based auth as the enforcement boundary" but no routes actually use these guards yet. Either wire them into the routes in this PR, or explicitly scope the PR to just add the utilities + tests and note that integration is a follow-up.
2.  get_user_course_role  crashes if  enrollment.role  is  None 
// python
role = enrollment.role.lower() if enrollment else None
If an enrollment record exists but  role  is  NULL  in the database,  .lower()  raises  AttributeError . Add a null guard:
// python
role = enrollment.role.lower() if enrollment and enrollment.role else None
3. SECRET_KEY fallback to file — stale key risk across workers
The  _get_or_create_fallback_secret_key()  function persists a generated key to  .flask_secret_key . If that file gets deleted or corrupted while the app is running, a new key is generated, invalidating all active sessions silently. Consider:
- Logging at  CRITICAL  level (not  WARNING ) when falling back
- Failing hard in production if  SECRET_KEY  isn't set (rather than silently falling back)
- The current warning says "Set SECRET_KEY in the environment for production deployments" but doesn't enforce it
**Medium Issues**
4.  require_authenticated()  doesn't verify the user still exists
It only checks  session.get("user_id") . If a user is deleted while their session cookie is still valid, the session is stale. Consider querying the DB:
// python
def require_authenticated():
    user_id = session.get("user_id")
    if not user_id:
        raise UnauthorizedError("Not authenticated")
    user = db.session.query(User).filter_by(id=user_id).first()
    if not user:
        session.clear()
        raise UnauthorizedError("Session expired")
    return user_id
This trades a DB round-trip for correctness. A lighter alternative is to just trust the session (acceptable for now) but document the trade-off.
5.  require_course_role  doesn't check if the course exists
If the course ID is invalid,  get_user_course_role  returns  None  (no enrollment found), and the function raises  ForbiddenError . This leaks that the course doesn't exist vs. the user isn't enrolled. Consider adding a course existence check or documenting that this is intentional.
6.  SESSION_COOKIE_SECURE  defaults to  false 
// python
SESSION_COOKIE_SECURE = os.getenv('SESSION_COOKIE_SECURE', 'false').lower() == 'true'
In production over HTTPS, session cookies without  Secure  flag can be intercepted via mixed content. The warning log is good, but consider defaulting to  true  when  FLASK_ENV=production  or adding a startup check.
**Minor / Nitpicks**
7. Frontend 401 interceptor redirects to  /  which may not be the login page
// javascript
window.location.assign("/");
If the app's login page is at a different route (e.g.,  /login ), this should redirect there instead. Verify the routing.
8.  get_user_course_role  uses  g  (request-scoped) cache — good pattern
The per-request cache via  g.setdefault("_course_role_cache", {})  avoids repeated DB queries within a single request. Nice touch. Just make sure  g  is properly torn down between requests (Flask handles this automatically).
9. File-based key persistence has a minor race condition
The  O_CREAT | O_EXCL  pattern is correct for the initial creation race. But if two processes read the file simultaneously after it's created, there's no issue since the content is stable. The  FileExistsError  fallback is correct. 
10. Test coverage looks good — 8 tests covering auth utilities, role checks, session presence. Consider adding a test for  get_user_course_role  with a  None  role field (edge case from issue #2).

DarioDeSilva:
Pushed another round addressing the remaining feedback on this PR (now scoped to just the auth foundation, per the split above).

**Blocking issues from 07-09:**
1. SECRET_KEY fallback race condition + file permissions — already fixed in an earlier round (`os.open` with `O_CREAT | O_EXCL`, mode 0600, falls back to reading the winner's key on `FileExistsError`). Added test coverage for all three cases (permissions, reuse, the race itself), since there wasn't any before.
2. `get_user_course_role` not distinguishing "course not found" from "not enrolled" — added a docstring explaining this is intentional: both cases already collapse to the same generic 403 in `require_course_role`, so distinguishing them would only add a way to enumerate course IDs through the error path.
3. `SESSION_COOKIE_SECURE` defaulting to false — now defaults to `true` when `SESSION_COOKIE_SAMESITE` is explicitly set to `None` and `SECURE` isn't set, per your first suggestion in that comment. Added tests for the default-true, default-false, and explicit-override cases.

**From 07-09 suggestions:** the `g`-scoped request-lifecycle caching you suggested was already implemented, but had no test. Added two: one confirming a second call in the same request doesn't re-query, one confirming the cache does not carry over into a second request (this one caught a real gap in how I'd first written it, using `test_request_context` nested inside an already-active app context doesn't actually give a fresh `g` the way two separate real requests would, fixed the test to use independent app contexts).

**From 07-13:** the in-memory caching concern is the same `g`-scoped cache, now covered by the test above.

Full backend suite: 306 tests passing. Everything else outstanding (get_my_enrollment_role/get_user_enrollments, migration SQL, self-promotion validation, sider.js/localStorage) lives in `course.py`/`assignment.py`/frontend files that are part of the TA-role-backend and TA-role-frontend PRs, not this one — I noted where each lands in my comment above.

Ready for another look whenever you have a chance.

SashankN7:
Overall: Solid foundation. The auth utilities, session-based login, SECRET_KEY fallback, CORS fix, and 401 interceptor are well-implemented.
**Medium Issues**
> 1.  get_user_course_role  crashes if  enrollment.role  is  None 
> // python
> role = enrollment.role.lower() if enrollment else None
> If an enrollment exists but  role  is  NULL ,  .lower()  raises  AttributeError . Fix:
> // python
> role = enrollment.role.lower() if enrollment and enrollment.role else None
> 2. SECRET_KEY fallback — stale key risk
> If  .flask_secret_key  is deleted/corrupted at runtime, a new key is generated, invalidating all active sessions silently. Log at  CRITICAL  level, or fail hard in production when  SECRET_KEY  isn't set.
> 3. PR description says "No route uses  require_course_role() " — but destructive endpoints like  delete_assignment  and  delete_submissions  in  assignment.py  have zero auth checks. Either guard those in this PR or explicitly scope the description to match what's actually unprotected after merge.

**Minor**
> 4. Test coverage for auth utils and config is good. Consider a test for  get_user_course_role  with a None role field.

DarioDeSilva:
Pushed a fix for each item from your last round.

1. `get_user_course_role` crashing on a `None` role — fixed exactly as you suggested (`enrollment and enrollment.role`), added a test for the null-role case.
2. SECRET_KEY fallback stale-key risk — `_get_or_create_fallback_secret_key` now reports whether it generated a fresh key or reused an existing one. `create_app` logs CRITICAL when a new key was generated (covers both first-run and the file being deleted/corrupted mid-run, since either way existing sessions are now invalid) and WARNING when it's just reusing what's on disk. Added tests for both log paths. Didn't go with failing hard in production since there's no production/environment flag in this codebase to key that off of, and CRITICAL logging was the option you offered as sufficient.
3. PR description overstating what's protected — reworded the "what this doesn't do yet" section. It now says plainly that the destructive/admin endpoints in the route files this PR doesn't touch have no auth check today and won't until the next PR, rather than the more neutral-sounding "pre-existing behavior." Kept it to that level of specificity rather than enumerating the individual endpoints in the PR description itself.
4. Test coverage for the null-role case — added, see #1.

Full backend suite: 309 tests passing.

SashankN7:
**Blocking / Security Issues**
1.  get_my_enrollment_role  still accepts  user_id  from client
// python
user_id = request.args.get("user_id")  # ← Still trusts client input
This endpoint can expose any user's role in any course. Fix: derive  user_id  from  session.get("user_id")  like the other guards.
2.  get_user_enrollments  still accepts arbitrary  user_id 
// python
user_id = request.args.get("user_id")  # ← Returns ALL enrollments for any user
Same issue - should use session-based auth or restrict to admins only.
3.  delete_submissions  skips auth when assignment not found
// python
assignment_for_auth = db.session.query(Assignment).filter_by(id=assignment_id).first()
if assignment_for_auth:  # ← Skipped when None, continues to delete without auth!
    enrollment = db.session.query(Enrollment)...
Fix: fail with 404 before the auth check is skipped.
**Medium Issues**
4.  sider.js  uses raw  fetch()  instead of  service()  helper
// javascript
fetch(`${process.env.REACT_APP_API_URL}/get_my_enrollment_role?...`)
Bypasses axios interceptors, error handling, and base URL configuration. Use the  service()  helper from  services/index.js .
5. Frontend still sends  requester_id  to backend (dead code)
Multiple frontend files still send  requester_id  as a parameter, but the backend now derives it from session. Remove dead code or add a comment explaining why it's ignored.
6.  rerun_submission_autograder  has no authorization check
Any logged-in user can rerun the autograder for any submission by guessing/iterating  submission_id . Add at minimum a session check and verify the requester owns the submission or is an instructor/TA.
**Minor Issues**
7. Missing test for  get_my_enrollment_role  when user is not enrolled
Add a test verifying the endpoint returns 403 for non-enrolled users.
8. TA role restrictions are undocumented
TAs cannot delete assignments/submissions, update course settings, change roles, or update AI settings. Document this in the PR description or code comments.
9.  accountPopoverContent.js  doesn't clear  courseRole  from React context on logout
Only clears localStorage, but stale  courseRole  in GlobalContext might briefly persist after logout.

DarioDeSilva:
Going through both your recent comments together, since I'd missed one.

**First, the 19:19:07Z comment** — I'd only been checking the single latest comment/review each round instead of everything since my last reply, so I missed this one until now. Sorry for the delay. Status on its 10 items:

1. `require_course_role` never called in any route — addressed in the PR description's "What this doesn't do yet" section.
2. `get_user_course_role` crashes on `None` role — fixed (previous round).
3. SECRET_KEY stale-key risk — moot now: the fallback-file mechanism this was about got dropped entirely when I reconciled with #312, which hard-fails at startup if `SECRET_KEY` isn't set instead. No fallback path left to go stale.
4. `require_authenticated()` doesn't verify the user still exists — documented as an intentional tradeoff rather than adding a DB round-trip to every call: cascade-delete (from #312) removes a deleted user's enrollments with them, so any `require_course_role()` call for that user_id already 403s downstream. No current route calls `require_authenticated()` without a follow-up role check. Docstring added in `e50f6a3`.
5. `require_course_role` doesn't distinguish course-not-found from not-enrolled — documented as intentional (previous round): both already collapse to the same generic 403, so distinguishing them would only add an enumeration path.
6. `SESSION_COOKIE_SECURE` defaults false — fixed (previous round): defaults to `true` when `SESSION_COOKIE_SAMESITE=None` is set explicitly, plus a startup warning otherwise.
7. 401 redirect to `/` may not be the login page — checked `App.js`: `/` is in fact where unauthenticated users get routed (line 91-92), so this is correct as written, not a bug.
8. `g`-scoped cache — no action needed, that was a positive note.
9. File-based key persistence race — moot, same as #3.
10. Test coverage suggestion (None-role case) — added (previous round).

**Second, the one you just left** — every item in this one is about a file that isn't actually part of #330's diff:

1, 2, 7. `get_my_enrollment_role` / `get_user_enrollments` trusting client `user_id`, and the missing not-enrolled test — all in `course.py`, which #330 doesn't touch. Already fixed this exact way (session-derived `user_id`, the test included) in the TA-role-backend PR I have queued up to open once this one merges.
3. `delete_submissions` skipping auth on missing assignment — `assignment.py`, same situation, already fixed (404s before the auth check now) in that same queued PR.
6. `rerun_submission_autograder` no auth check — `submission.py`, also not touched by #330. This one's actually already fixed on `main`, independently, by your own PR #312 (`_verify_course_staff` now guards it).
4, 9. `sider.js` raw `fetch()`, and `accountPopoverContent.js` not clearing `courseRole` from context — both frontend, both not in #330. #9 is already fixed in my queued frontend work. #4 is genuinely still open, I'll fix it when I open that PR.
5. Dead `requester_id` params sent from frontend — already cleaned up in the queued frontend work.
8. Undocumented TA restrictions — will document in the TA-role PR descriptions, since that's where the feature actually lands; #330 is auth plumbing only, there's no TA-specific behavior to describe here yet.

So nothing left unaddressed in #330 itself. The course.py/assignment.py/submission.py/frontend items are real and mostly already fixed, just sitting in the next two PRs in the sequence rather than this one. Full backend suite: 304/304 passing.

yczhangcc:
image
one backend test is failed

SashankN7:
waiting for the fix for the backend test before i review again

SashankN7:
Most of my earlier comments were stale — they targeted code in the queued PRs, not this one. Here's what's actually left for this PR:

**1. SECRET_KEY approach conflicts with main**
Main (via PR #312) now requires  SECRET_KEY  via env var and raises  RuntimeError  if missing — no fallback. This PR still has a file-based fallback ( _get_or_create_fallback_secret_key ). These two approaches will conflict on merge. Either:
- Update this PR to match main's strict approach (recommended), or
- Explicitly note the reconciliation plan in the PR description

**2.  FRONTEND_URL  vs  FRONTEND_ORIGIN  naming**
This PR uses  FRONTEND_URL  for the CORS env var. Main uses  FRONTEND_ORIGIN . These will also conflict on merge. Pick one name consistently

Once these two naming/approach mismatches with main are resolved, it'll be good to merge.

DarioDeSilva:
Hey Sashank, took a close look at this since I want to make sure it's actually resolved before pinging you again.

Checked `backend/api/__init__.py` and `backend/config.py` against what GitHub's Files changed tab is rendering for the current commit (`fb817b6`): the `SECRET_KEY` handling matches main's hard-fail approach (no fallback file anywhere), and `FRONTEND_ORIGIN` is what's used, not `FRONTEND_URL`. Also swept the whole repo tree at that commit for any leftover references and came up empty.

One thing I noticed: your last formal review is pinned to commit `d2390d7`, from before the PR got split. If that's the page you're looking at, it might be showing the older pre-split diff rather than the latest one. Would you mind refreshing to the latest commit and taking another look when you get a chance? Happy to walk through anything specific if it still looks off.

SashankN7:
Sorry about that, the PR with the current commit looks good to merge. The auth foundation is well-scoped, well-tested, and consistent with main. The only things to note:

1. require_authenticated()  trusts session without DB check — documented, acceptable for foundation
2. get_user_course_role  doesn't distinguish "course not found" vs "not enrolled" — documented, intentional
3. No route uses  require_course_role()  yet — by design, queued PRs handle that
4. 16 new tests covering auth utilities, config, and edge cases

SashankN7 (approved):
LGTM, added a comment in Conversation tab.

kiat:
Thanks!

PR #332, opened 2026-06-28
Use backend env file in Docker Compose

yczhangcc:
updates Docker Compose to read the backend environment file from:
env_file:
./backend/.env
Expected result:
Docker Compose should no longer ask for a root .env file.
The backend container should start using values from backend/.env.


kiat (approved):
Thanks!

PR #333, opened 2026-06-29
feat: Student AI Feedback with Prompt Selection, Usage Limits, and Memory

SashankN7:
## Summary

Implements Issue #328: Personalized AI-Powered Student Feedback System with Student Memory and Reference Solution Comparison.

### Issue #328 - Student-Side AI Feedback
- Student prompt picker with instructor-configured enabled prompts
- Per-student request tracking (ai_feedback_requests table)
- Max requests and wait-time enforcement (server + client)
- Remaining request count display and countdown timer
- Disable prompts when limit reached

### Future Work - Student AI Memory
- AI chat messages table stores conversation history
- Last 20 messages + coding_insights included in system prompt
- Personalized feedback using student history

### Future Work - Reference Solution Comparison
- Compare to optimal solution prompt
- Assignment description included in AI context
- Algorithmic improvement suggestions

### New Features
- 3 new default prompts: check_code_syntax, compare_to_optimal_solution, personalized_feedback
- Student-facing GET /assignments/<id>/prompts endpoint
- AI feedback status endpoint
- AI architecture documentation

## Testing
- 259 backend tests passing
- 9 frontend tests passing

## Database Migrations Required
After merging, run:
```bash
cd backend
PYTHONPATH=. flask db upgrade
```

This creates 2 new tables:
- ai_feedback_requests
- ai_chat_messages


DarioDeSilva (changes requested):
The infrastructure here is solid -- the limit/status logic is clean, the test coverage is good, and the session auth additions are well done. Four issues need addressing before this can merge.

**1. `/assignments/<id>/prompts` backend route is missing**

`normalize_feedback_prompts` is imported in `ai_feedback.py` but no route using it exists on this branch. The frontend's `getAssignmentPrompts()` calls `GET /assignments/{id}/prompts`, which will 404. `AIChatPanel` silently falls back to `setPrompts([])`, so students see "No prompts available" and the prompt-selection feature is completely non-functional.

The route needs to be added to `ai_feedback.py`:

```python
@ai_feedback.route("/assignments/<assignment_id>/prompts", methods=["GET"])
def get_assignment_prompts(assignment_id):
    assignment_obj = db.session.query(Assignment).filter_by(id=assignment_id).first()
    if not assignment_obj:
        raise NotFoundError("Assignment not found")
    if not assignment_obj.ai_feedback_enabled:
        return jsonify({"ai_feedback_enabled": False, "ai_feedback_prompts": []}), 200
    prompts = normalize_feedback_prompts(assignment_obj.ai_feedback_prompts)
    enabled = [p for p in prompts if p["enabled"]]
    return jsonify({"ai_feedback_enabled": True, "ai_feedback_prompts": enabled}), 200
```

**2. Inflated `user_message` stored in chat history (`code_editor.py` line ~794)**

At line 794, `user_message` is mutated to prepend the full prompt instruction:
```python
user_message = f"{prompt_config['prompt']}\n\n{user_message}"
```
Then at line 870, that inflated value is what gets stored in `ai_chat_messages`. The comment says "Store a concise version of the student message (not the full code dump)" but the prompt boilerplate is embedded in every stored message. On future turns, all of that gets fed back into the system prompt, growing the context with redundant text. Capture the original message before the mutation and store that instead:

```python
original_message = data.get("message")
# ... mutate user_message for the AI call ...
store_chat_message(student_id, assignment_id, "user", original_message, prompt_id)
```

**3. `record_feedback_request` not wrapped in try/except (`code_editor.py` line ~876)**

`store_chat_message` is wrapped so storage failures don't break the response. But `record_feedback_request` is bare. If its DB commit fails, the student gets a 500 even though the AI already replied and the messages were stored. On the next request, `check_feedback_limits` will count one fewer request than was actually made, allowing an extra request past the limit. Wrap it the same way the chat storage is wrapped:

```python
try:
    record_feedback_request(student_id, assignment_id, prompt_id)
except Exception as e:
    logger.warning(f"AI_CHAT: Failed to record feedback request: {e}")
```

**4. Duplicate migration ID with PR #312**

Both this branch and `fix/cascade-delete-user-relationships` (PR #312) define revision `c4d5e6f7a8b9` with the same `down_revision`. Whichever merges second will fail `alembic upgrade head` with a duplicate revision error. Since the AI table migrations chain off the cascade-delete revision, coordinate with PR #312 so only one branch defines it and the AI tables chain from it.

yczhangcc (dismissed):
Beside the issues Dario already mentioned, I noticed a few more things while testing:

1. Alembic currently reports multiple heads on this branch: `b3c4d5e6f7a8`, `c4d5e6f7a8b9`, and `f6a7b8c9d0e1`. This is related to the migration overlap Dario mentioned, but the impact is visible now: `flask db upgrade head` is ambiguous until the migration chain is cleaned up.

2. The prompt text may be duplicated in the AI call. The frontend sends `prompt.prompt` as the message, and then the backend prepends `prompt_config["prompt"]` again when `prompt_id` is present. So the model may receive the same instructor prompt twice.

3. Small carry-over from PR #312: the instructor delete dialog says courses, assignments, submissions, and grades will be deleted, but `courses.instructor_id` is `RESTRICT`. So deleting an instructor with courses should fail instead. The dialog or the behavior should be aligned.
4. also i notice now ollama is set as default model everytime even this is in another PR

SashankN7:
just fixed the changes, ready for another review

yczhangcc:
Fixed:
- GET /assignments/<assignment_id>/prompts now exists.
- Prompt duplication looks fixed: the frontend sends prompt_id, and prompt.prompt is only used for tooltip display.
- Inflated prompt storage looks improved: the backend sends api_user_message to the AI but stores the shorter user_message.
- Instructor delete dialog / RESTRICT behavior looks aligned now.

**Still needs work:
- Alembic still reports multiple heads: b3c4d5e6f7a8 and f6a7b8c9d0e**1.
image

- record_feedback_request(...) still appears to be called directly without try/except in code_editor.py. This could cause a 500 if request logging fails after the AI response is already generated.
- Minor security/design concern: the new prompts endpoint checks enrollment using student_id from query params. Since this is student-facing, deriving the student from the session would be safer.
-  also i notice now ollama is set as default model everytime even this is in another PR
image

DarioDeSilva:
This PR has merge conflicts with main — GitHub is showing it as CONFLICTING. Please rebase on main and resolve the conflicts before I do another review pass. Once that's clean I'll go through the remaining items.

yczhangcc:
please remove frontend/package-lock.json before merge , thanks

DarioDeSilva:
Thanks for the big push here. The security hardening (session auth, enrollment checks, FK cascades, mandatory `SECRET_KEY`, locked-down CORS, rate limiting) is solid and should land. But I think the core AI-feedback feature isn't actually wired up yet, so flagging before merge:

**Blocking: prompt selection, memory, and reference-solution comparison don't reach the LLM**

In `routes/code_editor.py::ai_chat`:
- `instructor_prompt_text` is resolved from the selected `prompt_id` (~line 921) but never used. `user_prompt` (line 929) is built from only `code` and `user_message`, and `_get_ai_chat_reply` always sends the static `AI_CHAT_SYSTEM_PROMPT`. So picking "Compare to optimal solution" vs. "Check code syntax" vs. any other prompt produces the same request to the model.
- `get_chat_history` is imported but never called. Chat history is written via `store_chat_message` but never read back, so there's no actual "memory" in the request.
- Assignment description and `coding_insights` aren't referenced anywhere in this file either.

Net effect: the prompt picker, personalization, and reference-solution comparison (the features this PR is named for) currently have zero effect on the AI's response. Only the usage-limit/tracking scaffolding is functionally connected. Worth wiring `instructor_prompt_text`, chat history, and assignment description into `user_prompt`/system prompt before merging, and adding a test that asserts on the actual content passed to `_get_ai_chat_reply` so this can't silently regress again.

**Minor**
- `GET /assignments/<id>/prompts` doesn't call `_verify_student(student_id)` like every other student route added in this PR does. An authenticated caller can pass an arbitrary `student_id` to probe course enrollment. Low severity but inconsistent with the pattern elsewhere.
- No backend tests for the new `/assignments/<id>/prompts` or `/ai_feedback_status` endpoints.
- Migration docstrings are stale copy-paste: `a2b3c4d5e6f7` says `Revises: 45b5cf6cc787` but `down_revision = "1bdb41066778"`; `c4d5e6f7a8b9` says `Revises: b2c3d4e5f6a7` but `down_revision = "b3c4d5e6f7a8"`. The actual chain is correct, just the comments are misleading.
- `run_code`'s rate limiter is an in-process dict, so it won't hold a global limit across multiple workers in production.
- `FRONTEND_ORIGIN` silently falls back to `localhost:3000` if unset. Unlike `SECRET_KEY`, a missing prod value fails silently rather than loudly.

yczhangcc:
image
image
it is not compiled on my side

DarioDeSilva:
Thanks for the follow-up. The compile fix (`canSubmitOnTime`/`canSubmitLate`) looks good.

On the AI wiring: this only closes part of the blocking item. `instructor_prompt_text` now reaches the model, but the other two pieces from my last review are still open:

- `get_chat_history` is imported but never called anywhere in `code_editor.py`. Messages are stored via `store_chat_message` but never read back into a request, so there's still no memory in the actual LLM call.
- Assignment description and `coding_insights` still aren't referenced anywhere in `code_editor.py`.

Since "memory" and "reference solution comparison" are both in the PR title/summary, I'd like those wired in (or the scope explicitly trimmed) before merge, plus a test that asserts on the content passed to `_get_ai_chat_reply` so this can't regress silently again.

DarioDeSilva (approved):
The wiring commit closes out the blocking item — verified locally, all 40 tests pass including the 4 new ones asserting on prompt content. Chat history, assignment description, and coding_insights all reach the LLM now, and reusing `_verify_student`'s return value to avoid the extra query is a nice touch.

This is good to merge from my side. Remaining items are non-blocking follow-ups, happy to have these in a separate PR:

- `GET /assignments/<id>/prompts` still doesn't call `_verify_student(student_id)` like the other student routes.
- No backend tests for `/assignments/<id>/prompts` or `/ai_feedback_status`.
- Migration docstrings (`Revises:` comments) don't match `down_revision` in a couple of files.
- `run_code`'s rate limiter is in-process only, won't hold a global limit across multiple workers in production.
- `FRONTEND_ORIGIN` silently falls back to `localhost:3000` if unset.

DarioDeSilva:
Follow-up on the non-blocking items — thanks for the quick turnaround, this all looks right and 64 tests pass locally. Two small notes, neither blocking (already approved):

- The migration docstring mismatches are still open: `a2b3c4d5e6f7` says `Revises: 45b5cf6cc787` but `down_revision = "1bdb41066778"`, and `c4d5e6f7a8b9` says `Revises: b2c3d4e5f6a7` but `down_revision = "b3c4d5e6f7a8"`. Purely cosmetic, fine as a fast-follow whenever.
- Commit message says "Add 10 backend tests" — looks like 7 were added. Not a problem, just flagging for accuracy.

DarioDeSilva:
Confirmed the migration docstring fix looks correct, everything is looking good now. Thanks for closing out all the review items.

kiat:
Could you please resolve the conflicts so that I can merge it!

kiat:
Thanks!

PR #334, opened 2026-07-01
feat: Instructor submission method selection per assignment

SashankN7:
## Summary

Adds instructor control over which submission methods (file upload and/or code editor) are available per assignment. Only the selected options are shown to students.

### Backend
- **`backend/api/models.py`**: Added `allow_file_upload` boolean column to `Assignment` model (default `True` for backward compatibility)
- **`backend/migrations/versions/e5f6a7b8c9d0_add_allow_file_upload.py`**: Alembic migration branching from `d4e5f6a7b8c9`
- **`backend/routes/assignment.py`**: Added server-side validation in `create_assignment` and `update_assignment` to prevent both submission methods from being disabled

### Frontend
- **`CreateAssignment.js`**: Replaced single `enable_code_editor` toggle with two toggles (Allow File Upload + Enable Code Editor). Uses `Form.useWatch` for reactive validation warning when both are disabled.
- **`assignmentSettings/index.js`**: Added submission method toggles in the assignment settings/edit page
- **`assignments/index.js`**: **Bug fix** — `selectedAssignment` was never set when opening the submission modal. Now properly passes both `allowFileUpload` and `enableCodeEditor` props to the modal.
- **`assignment_modal.js`**: Handles all 4 submission method combinations:
  - Both enabled → choice UI with separator
  - Only file upload → upload dragger only
  - Only code editor → editor button only
  - Neither enabled → warning alert

## Backward Compatibility

- `allow_file_upload` defaults to `True`, so existing assignments continue to work as before (file upload always available)
- Old assignments without the field are handled with `!== false` checks in the frontend

## Test Results

- Backend: 24/24 assignment tests ✅, 71/71 course tests ✅
- Frontend: 28/33 passing (5 failures are pre-existing axios Jest config issues, unrelated to this change)

### DIFF
--- a/backend/ai_feedback/integration.py
+++ b/backend/ai_feedback/integration.py
@@ -465,6 +465,40 @@ def get_structured_feedback_from_claude(api_key, prompt, model, temperature, pas
     return parse_feedback_json(raw_response, "Claude", past_insights)
 
 
+def get_structured_feedback_from_ollama(base_url, prompt, model, temperature, past_insights):
+    """Sends request to Ollama and parses JSON feedback."""
+    response = requests.post(
+        f"{base_url}/api/chat",
+        json={
+            "model": model,
+            "messages": [
+                {
+                    "role": "system",
+                    "content": CORRECTNESS_SYSTEM_PROMPT,
+                },
+                {
+                    "role": "user",
+                    "content": prompt,
+                },
+            ],
+            "options": {
+                "temperature": temperature,
+            },
+            "stream": False,
+            "format": "json",
+        },
+        timeout=60,
+    )
+
+    if response.status_code >= 400:
+        raise ValueError(f"Ollama API error: {response.text}")
+
+    data = response.json()
+    raw_response = data.get("message", {}).get("content", "").strip()
+
+    return parse_feedback_json(raw_response, "Ollama", past_insights)
+
+
 def update_submission_feedback(submission_id, ai_feedback_json, new_insights):
     """Updates the database with the AI feedback and new student insights."""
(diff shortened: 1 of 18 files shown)


yczhangcc (changes requested):
The manual UI flow for submission method selection looks mostly good. However, I found a few issues that should be fixed 

1. Submission method restrictions should also be enforced on the backend. Right now, the unavailable submission options are hidden in the student modal, but students may still be able to bypass the UI by directly calling the file upload endpoint or navigating to the code editor route. If allow_file_upload is false, the backend should reject file upload submissions. If enable_code_editor is false, the backend should reject code editor submissions.
2. The branch still appears to have two Alembic heads: e5f6a7b8c9d0 and f6a7b8c9d0e1. 
3. Frontend build completed with warnings. The Jest issue I saw seems like test fragility rather than broken manual UI behavior. The existing test selects switches by index, for example screen.getAllByRole("switch")[1]. Since this PR adds new switches before the AI feedback switch, such as Allow File Upload and Enable Code Editor, that index may now point to the wrong switch. It would be better to query the intended switch by label/name instead, for example screen.getByRole("switch", { name: /enable ai feedback/i }).
4. concern:  This branch also includes Ollama-related changes from another PR. I am not fully sure whether that Ollama PR has already been approved/merged separately. But for the frontend submission-method work, I think it is okay to continue assuming the Ollama-related frontend pieces will be updated/handled later if needed.

yczhangcc (approved):
LGTM, thank you!

kiat:
Excellent! Thanks!

Issue #335, opened 2026-07-04
When Editor Submission is used by the Student the autograder cannot find the submissoin


kiat:
When Editor Submission is used, the submitted file is named after the assignment name and it is not the name that autograder needs it. Autograder is configured to get an exact name and run with it. 

Two solution is possible
1. Name the file after the assignment
2. Find how the file has to be named from the autograder. 

Also, when the editor is used, the python file cannot be executed properly because it needs program argument to access the data.

PR #338, opened 2026-07-05
fix: allow students to edit submission filename in code editor (#335)

SashankN7:
## Issue

Fixes #335: When Editor Submission is used by the Student, the autograder cannot find the submission because the file is named after the assignment name, not the name the autograder expects.

## Problem

The code editor hardcoded the filename from the assignment name (e.g., `spiral_assignment.py`), but the autograder expects a specific filename (e.g., `spiral.py`). Students had no way to change the filename, causing autograder failures.

### `frontend/src/components/CodeEditor.js`

- Added editable filename input in the CodeEditor toolbar with click-to-edit UX
- Added `onFileNameChange` prop to CodeEditor component
- Added `EditOutlined` icon import and `Input` component import from antd
- When clicked, the filename becomes an editable input field; pressing Enter or blurring saves the change
- Added tooltip explaining filename importance for autograder compatibility

### `frontend/src/pages/codeEditor/index.js`

- Added `onFileNameChange` handler that updates filename state and saves the draft
- Fixed stale filename bug in `saveDraft` by using `fileNameRef` instead of closure variable
- Ensured filename changes are persisted immediately so the correct filename is used on submission

## How to Test

1. Open an assignment with code editor enabled
2. Notice the filename displayed in the editor toolbar (e.g., "spiral_assignment.py")
3. Click the filename — it becomes an editable input field
4. Change the filename to match the autograder's expected name (e.g., "spiral.py")
5. Press Enter or click outside — the filename is saved
6. Submit the assignment — the autograder should now find the correct filename

### DIFF
--- a/frontend/src/components/CodeEditor.js
+++ b/frontend/src/components/CodeEditor.js
@@ -1,5 +1,5 @@
 import React, { useRef, useEffect, useState } from "react";
-import { Button, Space, Tag, Tooltip } from "antd";
+import { Button, Input, Space, Tag, Tooltip } from "antd";
 import {
   SaveOutlined,
   PlayCircleOutlined,
@@ -9,6 +9,7 @@ import {
   CheckCircleFilled,
   ClockCircleOutlined,
   CaretRightOutlined,
+  EditOutlined,
 } from "@ant-design/icons";
 
 // CodeMirror 6 imports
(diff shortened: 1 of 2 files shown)


kiat:
Thanks!

Issue #336, opened 2026-07-04
When Editor Submission is used, there is no way for students to resubmit!


kiat:
After submission, there is no way to go back to the editor again and resubmit.

PR #339, opened 2026-07-05
fix: allow students to resubmit via code editor after submission (#336)

SashankN7:
## Issue
Fixes #336: When Editor Submission is used, there is no way for students to resubmit!

## Problem
After submitting via the code editor, students were taken to the results page with no way to go back to the editor and resubmit. The existing 'Resubmit' button only opened the file upload modal, not the code editor.

## Changes

**frontend/src/pages/result/ActionButtons.js:**
- Added CodeOutlined icon import
- Added onResubmitEditor and enableCodeEditor props
- When enableCodeEditor is true, shows 'Resubmit via Editor' button instead of the file upload resubmit

**frontend/src/pages/result/index.js:**
- Added useNavigate import
- Added enableCodeEditor state variable
- Fetches enable_code_editor from assignment details
- Passes onResubmitEditor handler that checks due dates before navigating to /codeEditor/:assignmentId
- Passes enableCodeEditor to ActionButtons

## How It Works
When the assignment has code editor enabled, the results page now shows a 'Resubmit via Editor' button instead of the file upload resubmit button. Clicking it navigates the student back to the code editor with their previous code and draft loaded, allowing them to make changes and resubmit.

### DIFF
--- a/frontend/src/pages/result/ActionButtons.js
+++ b/frontend/src/pages/result/ActionButtons.js
@@ -1,7 +1,7 @@
 import { Button, Space } from "antd";
-import { ReloadOutlined, UploadOutlined, DownloadOutlined, ClockCircleOutlined } from "@ant-design/icons";
+import { ReloadOutlined, UploadOutlined, DownloadOutlined, ClockCircleOutlined, CodeOutlined } from "@ant-design/icons";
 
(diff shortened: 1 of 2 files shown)


kiat:
Thanks!

Issue #337, opened 2026-07-04
AI is not configured for this course.


kiat:
When Gemini is used, then I get this error all the time. 

AI is well configured and can refresh the models but it complains that there is no AI settings for the course. 

Also, the customization for the assignment needs to include the Keys again because it can be different than the course level settings.

yczhangcc:
solved

Issue #12, opened 2023-01-25
Create course


Nils19:
Creating a course does not go through, Operation failed. Fix issue

Nils19:
Incorporate adding students to this course.

Nils19:
instruct_id instead of instructor_id index.js

PR #342, opened 2026-07-06
fix: resolve course AI config, Gemini timeouts, and editor resubmission bugs (#336, #337)

yczhangcc:
#### 1. Fixed "AI is not configured for this course" Errors (#337)
* **The Problem:** The app crashed when trying to generate feedback with Gemini or other providers, incorrectly claiming the course had no credentials even when keys were validly refreshed. Assignment-level custom keys were ignored, forcing strict compliance with course-level defaults.
* **The Fix:** 
  * Extracted credential lookup logic into a dedicated helper `get_assignment_provider_credential()`.
  * Refactored `get_provider_credentials()` to pass the `assignment` entity through and evaluate `assignment.use_course_ai_default`. 
  * Updated all provider routing blocks (`openai`, `gemini`, `claude`, `ollama`) to intercept and prioritize decrypted assignment-specific keys over course defaults when configured.
  * Exposed `ai_feedback_api_key` in `serialize_assignment_ai_settings` to ensure configuration fields accurately serialize to the UI layer.

#### 2. Implemented Gemini API Transient Error Recovery (#337)
* **The Problem:** Direct `requests.post` network calls to Gemini were brittle, triggering persistent failures for users when hitting rate limits or minor provider-side service interruptions.
* **The Fix:** Implemented a robust wrapper `post_gemini_with_retry()`. This enforces up to 3 retries using a linear backoff strategy (`attempt * 1s`) when intercepting specific server-side/network response status codes (`429`, `500`, `502`, `503`, `504`).

#### 3. Resolved Editor Resubmission Lockout (#336)
* **The Problem:** Once a student completed an initial assignment attempt using the online code editor, the UI locked them out, completely preventing any secondary attempts or iterative resubmissions.
* **The Fix:** Adjusted the frontend navigation state constraints to ensure that after initial code submissions are recorded, the workspace properly unblocks access to the editing suite for continuous student resubmissions.

#### 4. Streamlined Testing Scripts
* **The Problem:** Frontend test targets inside the `Makefile` were left hanging or polling for file watching triggers, causing continuous integration issues.
* **The Fix:** Configured the `test-frontend` target explicitly with `--watchAll=false` and `--silent` flags to guarantee tests complete deterministically in headless environments.


SashankN7 (changes requested):
**Critical**
1.  get_provider_credentials  doesn't validate provider matches the credential origin
 backend/ai_feedback/integration.py  — In the new  get_provider_credentials , if  assignment_credential  is set, it's used directly without verifying it belongs to the correct provider. An instructor could configure an assignment with  ai_feedback_provider = "gemini"  but paste an OpenAI API key into the assignment key field. The code would blindly use the OpenAI key against Gemini's API, producing a confusing 401/403 error with no clear explanation.
// python
# Current — uses whatever credential is set regardless of provider
if assignment_credential:
    api_key = assignment_credential
else:
    ...
Suggestion: Either (a) store the credential keyed by provider (e.g., a JSON field mapping provider → encrypted key), or (b) at minimum, log/warn when the assignment credential doesn't match the resolved provider.
2.  _get_ai_chat_reply  does NOT validate Ollama URLs — SSRF risk
 backend/routes/code_editor.py  — The feedback path ( get_structured_feedback_from_ollama ) calls  validate_ollama_url(base_url)  before making requests. The new  _get_ai_chat_reply  omits this check for Ollama:
// python
if provider == "ollama":
    response = requests.post(
        f"{api_key}/api/chat",  # <-- no validate_ollama_url() call
This is the exact SSRF vulnerability that was mitigated in the  ai_feedback  module. Both paths should enforce URL validation.
**Suggestions**
3.  post_gemini_with_retry  only retries on status codes, not on network exceptions
 backend/ai_feedback/integration.py  —  requests.post  can raise  requests.ConnectionError ,  requests.Timeout , etc., which are common transient failures. These bypass the retry loop entirely. Consider wrapping the  requests.post  in a try/except inside the loop:
// python
for attempt in range(1, GEMINI_MAX_ATTEMPTS + 1):
    try:
        response = requests.post(...)
    except (requests.ConnectionError, requests.Timeout):
        if attempt == GEMINI_MAX_ATTEMPTS:
            raise
        time.sleep(GEMINI_RETRY_BACKOFF_SECONDS * attempt)
        continue
    if response.status_code not in GEMINI_TRANSIENT_STATUS_CODES:
        return response
    ...
4.  post_gemini_with_retry  has no logging on retry
Hard to diagnose in production when retries are silently happening. Add at least a  print()  (matching the project's logging convention):
// python
if attempt < GEMINI_MAX_ATTEMPTS:
    print(f"GEMINI_RETRY: status {response.status_code}, attempt {attempt}/{GEMINI_MAX_ATTEMPTS}", flush=True)
    time.sleep(GEMINI_RETRY_BACKOFF_SECONDS * attempt)
5.  post_gemini_with_retry  isn't used for  get_structured_feedback_from_gemini 
The feedback path still uses a bare  requests.post . Only the chat endpoint uses the retry wrapper. If Gemini rate-limiting is the root cause (as the PR description states), the feedback path is equally (or more) affected. Apply the retry wrapper to  get_structured_feedback_from_gemini  as well.
6.  get_structured_feedback_from_gemini  should be updated to use  post_gemini_with_retry 
The diff shows the function was refactored but the main feedback path still uses  requests.post  directly:
// python
response = requests.post(
    f"https://generativelanguage.googleapis.com/...",
    params={"key": api_key},
    json={...},
    timeout=30,
)
This is the primary Gemini call path. Apply the same retry wrapper here for consistency.
7. Linear backoff ( attempt * 1s ) is conservative for 429s
Google recommends exponential backoff with jitter for 429 responses. Consider at minimum  2 ** (attempt - 1)  or follow Google's own recommended pattern.
8. Error message conflates "course not found" with "AI not configured"
 backend/routes/code_editor.py :
// python
if not course:
    raise BadRequestError("AI is not configured for this course. Please ask your instructor to set up an API key.")
This should say  "Course not found for this assignment."  (or similar). The course-not-found case is a data integrity issue, not an AI configuration issue.
9.  package-lock.json  is +20K lines but  package.json  isn't in the diff
The lockfile change suggests a full regeneration (different npm/Node version). This inflates the diff significantly and risks unrelated dependency drift. If no new packages were intentionally added, consider regenerating the lockfile on the same Node version as  main  and regenerating only the changed entries, or excluding the lockfile from this PR if no dependencies changed.
10.  ai_feedback_api_key  is cleared when  use_course_ai_default  is toggled, but the frontend UX doesn't make this clear
 backend/ai_feedback/settings.py :
// python
if assignment.use_course_ai_default:
    assignment.ai_feedback_api_key = ""
The frontend shows an Alert ("Paste a new value only when replacing it") but doesn't indicate that toggling "Use course defaults" will permanently delete the saved key. This could lead to accidental data loss. Consider either (a) soft-clearing (nulling the key but keeping it in the DB for reuse) or (b) showing a confirmation dialog.
11.  ai_feedback_api_key  column uses  nullable=False  default in model but the migration uses  DEFAULT '' 
 backend/api/models.py :
// python
ai_feedback_api_key = db.Column(db.String, default="")
 backend/migrations/versions/f7a8b9c0d1e2 :
// python
op.execute("ALTER TABLE assignments ADD COLUMN IF NOT EXISTS ai_feedback_api_key VARCHAR DEFAULT ''")
These are consistent, but the  default=""  in the SQLAlchemy model means every new Assignment object gets  ""  (not  NULL ), and the migration defaults to  "" . This is fine — just noting that the encrypted key will be stored as  ""  when not set, and the  get_assignment_provider_credential  function correctly handles this with  .strip() .
12.  onResubmitEditor  prop is in  ActionButtons  but the  result/index.js  caller doesn't pass it
In  frontend/src/pages/result/index.js ,  ActionButtons  is rendered but the  onResubmitEditor  prop is not passed. This means the "Resubmit via Editor" button renders but clicking it calls  undefined . The  enableCodeEditor  state also isn't wired from the assignment data in this file (it's read from  res.data.enable_code_editor  on line ~128 of the diff, but the state variable declaration and setter need to be present).
Looking more closely at the diff for  result/index.js , I see  enableCodeEditor  state is declared and  setEnableCodeEditor  is called, but  onResubmitEditor  is never passed to  ActionButtons . The button will render but be non-functional. Please wire the handler or remove the prop.
13.  ActionButtons.test.jsx  — test for "same default style" asserts  not ant-btn-primary  but doesn't verify both buttons do render
The third test case ("uses the same default style") passes when both buttons render, but only checks the negative case. Consider adding  toBeInTheDocument()  assertions first to guard against false positives from  queryByRole  returning null.
**Minor Considerations/Fixes**
14. Merge migration  1bdb41066778  has  down_revision = ('d4e5f6a7b8c9', 'f7a8b9c0d1e2')  — verify these are the correct heads
Ensure the migration graph is clean. If  f7a8b9c0d1e2  depends on  e5f6a7b8c9d0  and  d4e5f6a7b8c9  is the other head, this is correct. Just double-check with  flask db heads  and  flask db check .
15.  _get_ai_chat_reply  uses  timeout=30  for all providers, but  get_structured_feedback_from_ollama  uses  timeout=60 
For consistency, Ollama in the chat path should probably also use 60s timeout.
16.  CreateAssignment.js  —  fetchAiModels  is called with  assignmentApiKey  in the payload, but the backend  test_ai_model  endpoint needs to accept that key
Ensure the  /test_ai_model  route has been updated to accept an optional  api_key  field in the request body. If not, the custom key will be silently ignored when fetching models.
17. Test file  assignments/index.test.jsx  removes the assertion  expect(screen.getByTestId('assignment-modal')).toHaveTextContent('OPEN') 
This reduces test coverage for the modal open behavior. If this was removed because the assertion was wrong/flaky, consider replacing it with a corrected assertion rather than removing it entirely.
18. Commit messages are not descriptive enough
 "fix gemini and ai setting"  and  "merge latest main into Gemini AI settings fix"  are not conventional commit messages. Consider squashing and rewriting with more context, e.g.:
-  feat: add assignment-level AI API key support 
-  fix: apply Gemini retry logic to all request paths 
-  fix: wire file-upload and editor resubmit buttons independently 
Summary
The core changes (assignment-level AI credentials, Gemini retry, editor resubmission fix, Makefile CI fix) are well-structured. The main concerns are:
1. SSRF gap in  _get_ai_chat_reply  Ollama path (#2)
2. Provider-credential mismatch risk (#1)
3. Missing retry wrapper on the primary feedback path (#5, #6)
4. Dead  onResubmitEditor  button in result page (#12)
5. Large lockfile diff adding noise (#9)

yczhangcc:
hi i have update and fixed all the issues ,thanks
1.Fixed provider key mismatches, added Ollama SSRF chat validation, locked timeout at 60s, and implemented Gemini exponential backoff retries 
2.Corrected "course-not-found" error messages, added key-clearing warning prompts to assignment UIs, and verified "Resubmit" editor component wiring
3.Resolved local Windows PermissionError workspace crashes by isolating pytest temp paths, fixed ActionButtons render assertions, and restored modal test paths 
4.Removed local package-lock.json noise from Git tracking, leaving all local test suites passing cleanly with a stable 63% global code coverage baseline
5.fix Rerun Aotugrade missing issue #335

kiat:
Thanks!

PR #346, opened 2026-07-11
Fix AI settings detail format and assignment date UI

yczhangcc:
## Summary
- Remove seconds from assignment date/time pickers
- Move Test Selected Model button after Default Model selection
- Hide Late Due Date unless Allow Late Submissions is enabled
- Prevent stale late due date values from being submitted
- Add frontend tests for the updated behaviors

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ test: test-backend test-frontend  ## Run all tests
 
 test-backend:                     ## Backend unit / integration tests
 	@echo "Running backend tests..."
-	@cd ./backend && python3 -m pytest
+	@cd ./backend && python -m pytest
 	@echo "Backend tests passed."
 
 test-frontend:                    ## Front-end Jest tests
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,14 @@
+ifeq ($(OS),Windows_NT)
+PYTHON ?= python
+else
+PYTHON ?= python3
+endif
+PIP ?= $(PYTHON) -m pip
+
 .PHONY: install test clean
 install:
 	@echo "Installing backend dependencies..."
-	@pip install -r ./backend/requirements.txt
+	@$(PIP) install -r ./backend/requirements.txt
 	@echo "Backend dependencies installed."
 	@echo "Installing frontend dependencies..."
 	@cd ./frontend && npm install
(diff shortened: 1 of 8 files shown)


DarioDeSilva:
Reviewed this and it looks solid. Verified the two behavioral fixes directly:

- The stale late due date bug: nulling `late_due_date` on submit when `allowLateSubmissions` is false is correct given the field unmounts, otherwise Antd's form store would keep resubmitting the old value. Confirmed the edit page already had this guard and the create page now matches it.
- The date pickers no longer show seconds, confirmed via the shared `format`/`showTime` props.
- Test Selected Model now sits next to the model selector as described.

Two small non-blocking notes, feel free to ignore for this PR:

1. `dateTimePickerProps` is duplicated identically in both `assignmentSettings/index.js` and `CreateAssignment.js`. Could live in a shared constants file at some point.
2. The late due date conditional render was applied to both the create and edit assignment pages, but only the create page got a test for it.

Tests are meaningful, not just mocked past the logic, and CI is green. This is ready to merge from my side.

DarioDeSilva (approved):

SashankN7 (changes requested):
No blockers here, but just flagging these (and requesting changes) because they should be addressed before we can merge:

**Comment 1:**
> The Makefile  test-backend  target was changed from  python3  to  python . On many systems (especially fresh Ubuntu/Debian installs),  python  either doesn't exist or points to Python 2. This will break the test suite. Recommend reverting to  python3  or using  $(PYTHON) .

**Comment 2:**
>  dateTimePickerProps  is defined identically in both  assignmentSettings/index.js  and  CreateAssignment.js . Extract it to a shared constants file so it stays DRY.

**Comment 3:**
>  CreateAssignment.js  correctly sets  late_due_date  to  null  when late submissions are disabled. Does  assignmentSettings/index.js  do the same in its PUT payload? If not, a stale  late_due_date  will persist — this could be a real bug.

SashankN7 (approved):
**1. Makefile python3 → python — FIXED**
> Nice fix — using  $(PYTHON)  with cross-platform detection ( python3  on Unix,  python  on Windows) is the right approach.

**2. dateTimePickerProps extracted to shared file — FIXED**
> Extracted to  frontend/src/constants/dateTimePicker.js  as  ASSIGNMENT_DATE_TIME_PICKER_PROPS . Both  assignmentSettings/index.js  and  CreateAssignment.js  now import from there. Clean DRY fix.

**3. late_due_date null-setting verified — FIXED**
> New test in  assignmentSettings/index.test.jsx  confirms that disabling late submissions submits  late_due_date: null . This was the bug I was worried about — good that it's covered now.

**4. New test coverage added**
> The new test suite covers the late due date field visibility toggle and the null-setting behavior. This is solid test coverage for the UI change.

kiat:
Thank you everyone!

PR #347, opened 2026-07-13
Replace per-submission container startup with persistent containers

RaymondZhu3:
# Replace per-submission container startup with persistent containers

Addresses #87

## Summary

Replaces the per-submission ephemeral Docker container with a persistent container reused across all submissions to the same assignment. Container is created once when the autograder is uploaded (or on first submission) and reused for subsequent submissions instead of being created and torn down every time.

## Problem

Every student submission spun up a new Docker container and destroyed it after grading. Measured overhead:

| | Time |
|---|---|
| Container creation | ~0.81s |
| Container teardown | ~1.43s |
| **Total per-submission overhead** | **~2.24s** |

This overhead was paid on every single submission, regardless of whether the assignment's autograder image had already been used.

## Solution

- Container is created once per assignment (on autograder upload, or lazily on first submission) and left running.
- Subsequent submissions reuse the existing container via `assignment.container_id`, looked up from the database.
- Edge cases handled:
  - Stored `container_id` is stale (container was removed) → falls back to lookup by name, then recreates if needed.
  - Concurrent requests race to create the same assignment's container for the first time → retry loop catches the resulting `409 Conflict` / `404 Not Found` and recovers.
  - Autograder image changes after container creation → detected via image tag comparison, container is recreated.

## Metrics

All trials run locally, 5 repeats per scenario unless noted, `--max_execution_time 1`.

### Overall Summary

| Scenario | Ephemeral | Persistent | Speedup |
|---|---|---|---|
| Single submission (Sub 1) | 6.79s | 5.01s | ~26% faster |
| Sub 2 (warm) | 6.78s | 4.41s | ~35% faster |
| Sub 3 (warm) | 7.21s | 4.39s | ~39% faster |
| Concurrent (3 workers) | 7.53s | not measured | see Known Limitations |

### Warm-up / Degradation Trend

| Approach | Sub 1 | Sub 2 | Sub 3 | Trend |
|---|---|---|---|---|
| Ephemeral | 5.89s | 6.78s | 7.21s | ↑ +22% |
| Persistent (subprocess) | 4.54s | 4.41s | 4.39s | → flat |

Ephemeral submission time increases across repeated submissions within a session, consistent with accumulating Docker daemon overhead from repeated container creation/teardown. Persistent containers avoid this as timing stays relatively constant regardless of submission count.

### Container Lifecycle Overhead (root cause of the gap)

| | Startup | Teardown | Total Overhead |
|---|---|---|---|
| Ephemeral (every submission) | 0.81s | 1.43s | 2.24s |
| Persistent, Sub 1 (creates once) | ~0.3–0.5s | 0s (never torn down) | ~0.3–0.5s |
| Persistent, Sub 2+ (reused) | ~0.1–0.2s | 0s | ~0.1–0.2s |

Initial data showed persistent Sub 1 was faster than ephemeral Sub 1 despite both creating a new container which was unexpected, since creation cost should be similar either way. Experimentation showed that creation time is comparable between the two approaches, but ephemeral additionally pays teardown cost (`container.stop()` + `container.remove()`) on every submission, including the first. Persistent containers never pay this cost since the container stays alive. Teardown, not creation, is the larger and more consistent component of the overhead ephemeral pays.

### Where the remaining time goes

Container lifecycle (startup+teardown) was never the dominant cost. Actual autograder execution takes ~4s regardless of container approach (tested with a 1s intentional delay plus real test execution). This is the primary bottleneck and a possible target for optimization.

## Known Limitations

**Concurrent submissions to the same assignment can corrupt each other's results.**

If multiple students submit to the *same* assignment at nearly the same time, they share one persistent container's filesystem. Two issues were found and investigated:

1. **Container-name creation race** (fixed in this PR): concurrent first-time submissions to a brand-new assignment could all attempt to create a container with the same name, causing a `409 Conflict`. Fixed via a retry loop that falls back to looking up the container that won the race.
2. **Shared results file corruption** (not fixed in this PR): concurrent submissions write to and read from the same fixed path (`/autograder/results/results.json`) inside the shared container. One submission's write can interleave with another's read, corrupting the JSON. Found via stress testing (`--max_workers 3`).

Proposed fix for (2):
- Move to one persistent container per worker (paired with a Celery/Redis task queue) rather than one container per assignment, so concurrent submissions never share container state at all. This also solves the throughput problem of many simultaneous submissions.

This limitation does not affect the ephemeral (main) approach, since each submission there gets an isolated container.

## Changes

- `backend/api/models.py` — `container_id` field on `Assignment` (uncommented/re-enabled)
- `backend/routes/submission.py`:
  - `get_or_create_assignment_container()` — new function implementing container reuse, stale-ID fallback, image-change detection (handles Docker's tag deduplication across identical-content images), and a retry loop for concurrent container-creation races
  - `reset_assignment_container()` — discards a persistent container after a stuck/timed-out run
  - `upload_submission()` — uses persistent container instead of ephemeral create/destroy
- `backend/migrations/env.py` — fixed a pre-existing bug where `target_metadata` was correctly set to `db.metadata`, then immediately overwritten to `None` by leftover Alembic template boilerplate a few lines later which broke `flask db migrate`.
- `backend/migrations/versions/a3f92c1e7b04_add_container_id_to_assignments.py` — new migration adding the `container_id` column (previously only present via manual `ALTER TABLE`, with no migration tracking it)

## Testing

Stress-tested via `backend/test/stress/test_long_running_grader.py` across:
- Single submission (`--num_submissions 1 --max_workers 1`)
- Sequential submissions (`--num_submissions 3 --max_workers 1`)
- Concurrent submissions (`--num_submissions 3 --max_workers 3`)

All Ephemeral scenarios and sequential Persistent scenarios (single submission, sequential submissions): 100% pass rate across 5 trials each. Concurrent Persistent submissions (3 workers) failed consistently due to the results-file corruption.


DarioDeSilva:
Will review as soon as conflicts are merged!

SashankN7 (changes requested):
**Critical / Blocker**
1. Concurrent submission filesystem corruption — needs a mitigation before merge
> The PR body acknowledges that concurrent submissions to the same assignment share  /autograder/results/results.json  and can corrupt each other's results. Even though the full fix (one container per worker) is deferred, I think we need at least a basic mitigation before merging — e.g., writing to a unique temp path per submission ( /autograder/results/results_{submission_id}.json ) and reading from that path after the run, instead of the fixed  results.json . Without this, two students submitting at the same time can get garbled grades. Can you address this?
2. No cleanup when an assignment is deleted — container resource leak
> Persistent containers are never stopped/removed when an assignment is deleted from the database. Over time this will leak Docker containers on the server. The  Assignment  model needs either a SQLAlchemy  after_delete  event, a cascade, or a background cleanup task that stops containers for assignments that no longer exist. Can you add this?

**Important / Should Fix**
3.  container.reload()  can throw uncaught exceptions
> If the container is removed between  containers.get()  and  container.reload() , this will raise an uncaught  NotFound  or  APIError . Wrap the image-change detection block in a try/except:
> // python
> try:
>     container.reload()
>     # ... image tag comparison ...
> except docker.errors.NotFound:
>     container = None
4. Cleanup command ( rm -rf /autograder/... ) has no error checking
>  container.exec_run("sh -c 'rm -rf /autograder/submission/* /autograder/results/*'")  returns a result but the exit_code is never checked. If this fails silently (e.g., permission issue, container state issue), the submission could get contaminated results from a previous run. Should check  exit_code  and raise/abort if it fails.
5.  container.stop()  /  container.remove()  in image-change path has no error handling
> In the image-mismatch branch, if  container.stop()  or  container.remove(force=True)  throws, the exception propagates unhandled. Wrap in try/except and fall through to recreation:
> // python
> try:
>     container.stop()
>     container.remove(force=True)
> except Exception:
>     pass
> container = None
6. Debug print statements should be proper logging
>  print(f"[DEBUG] unexpected error getting container: {e}", flush=True)  — use  logging.warning()  or  logging.error()  instead of print with  [DEBUG]  prefix. This is production code and debug prints shouldn't ship.

**Minor / Nice to Have**
7.  import time  is unused
>  time  is imported but never referenced anywhere in the file. Remove it.
8.  _normalize_tag  is a nested function recreated on every call
> It's defined inside  get_or_create_assignment_container  but has no dependency on the outer scope. Move it to module level (or as a static helper) so it's only defined once:
> // python
> def _normalize_tag(name):
>     if name and ':' not in name:
>         return name + ':latest'
>     return name
9. Migration file missing trailing newline
>  a3f92c1e7b04_add_container_id_to_assignments.py  — the last line ( op.drop_column(...) ) has no trailing newline. Minor but convention is to end files with a newline.
10.  container_name  variable shadowing is slightly confusing
> In  upload_submission() ,  container_name = container.name  is set but only used in the  subprocess.run   docker exec  call. Meanwhile,  container.exec_run()  is used directly for other operations. Consider using  container.exec_run  consistently for the autograder execution too (via  subprocess  inside the container isn't needed — you already have the Docker SDK container object).
11.  db.session.commit()  inside  get_or_create_assignment_container  — consider the caller's context
> This function commits the DB session to save  container_id . If the caller later needs to roll back (e.g., submission fails), the  container_id  commit is already persisted. Consider whether this should be left to the caller or if you need  db.session.flush()  + commit at the end of the request.

RaymondZhu3:
Merge conflicts should be fixed now and requested changes have been addressed. Ready for another review.

yczhangcc:
Hi , Thanks for the work, the persistent container approach improves performance by avoiding repeated container startup and teardown overhead.
A few implementation details need to change and consider :
1. In `upload_submission()`, please verify that the old temporary container creation/removal code is completely removed. Otherwise, the code may still create a new container per submission and lose the performance benefit.
2. For result isolation, instead of running the autograder and then moving `results.json`, consider passing a unique output path or submission-specific results directory before execution. Multiple submissions can still write to `results.json` at the same time before the rename happens.
3. In `get_or_create_assignment_container()`, consider adding more explicit cleanup/error handling around container creation failures. For example, if container creation succeeds but saving `container_id` fails, the container could become orphaned.
4. For scalability, the current one-container-per-assignment design may still limit concurrent submissions because students share the same container filesystem. A future worker-based container pool could provide better isolation and throughput.

same time, i just want to offer some possible **future** improvements for scalability:
1. Container management: Consider moving from one persistent container per assignment to a worker-based container pool. This could improve concurrency and avoid multiple submissions sharing the same filesystem.
2. Async grading pipeline: For large classes, a queue-based grading system (e.g., Redis + Celery workers) could prevent long-running grading jobs from blocking API requests and allow better workload distribution.
3. Resource management: Add monitoring and automatic cleanup/recycling strategies for idle containers to prevent resource usage growth over time.
4. Performance optimization: Consider optimizing file transfer and execution overhead (e.g., volume mounts or more direct Docker SDK usage) if grading becomes a bottleneck.
These may **not be required for the current implementation**, but could be useful directions as CodeAssist scales to larger courses.
And also previously i have finished the stress test part, so when you do the improvement , you maybe also able to use that to test the time.   **the path is backend/test/stress**

RaymondZhu3:
Thanks for the review. Here's what I fixed:

1. Old temporary container create/teardown code in `upload_submission()`: confirmed removed. Verified that the same container persists and gets reused across multiple submissions to an assignment. No new container is created or torn down per submission.
2. Currently working on addressing this issue using filesystem isolation for each submission.
3. `get_or_create_assignment_container()`: `container.start()` failures now log and raise instead of failing silently. `db.session.commit()` failures now roll back, stop/remove the orphaned container, log, and raise instead of returning a stale reference.
4. Agreed, planning to move grading to a Celery/Redis worker pool for long term scalability where Flask enqueues a task instead of grading inline, and a pool of workers pull tasks and run them in parallel across separate containers. The container get/exec logic from this PR should mostly carry over into each worker's task.

Also discovered that deleting all assignments in a course wasn't cleaning up their containers because the cleanup only fired on single-assignment deletes. Added cleanup for bulk deletion too.

Future-improvements list all makes sense as follow-on work and I'm planning on working with Khushi to make these improvements.

Also I ran some stress tests, but the ones that call `upload_submission` and cleanup's `delete_user` fail with 403s, since those two routes now require auth.

DarioDeSilva (changes requested):
Went through this in depth, including checking the actual branch code (not just the diff) against Sashank's review and yczhangcc's comments.

Good news: items 2 through 9 and yc-3 (container cleanup on delete, reload() exceptions, exit_code checks, stop/remove error handling, debug prints, unused import, _normalize_tag placement, migration newline, orphaned container on commit failure) all check out fixed. Verified the delete_all_assignments bulk-delete path specifically since it bypasses the ORM after_delete event, and the explicit cleanup loop there covers it correctly. Also verified the new migration is a single clean head on top of c4d5e6f7a8b9 with no branching.

One item I don't think is fully resolved yet: #1, the concurrent-submission corruption blocker.

The results filename fix (results_{submission_uuid}.json, renamed after run_autograder finishes) protects the final artifact name, but it's still the "move after" pattern both of you flagged rather than a unique path used during execution. Two submissions to the same assignment can still overlap on the shared results.json path while run_autograder is running, before either mv fires.

It's also a bit worse than before in one respect: there's no locking anywhere in upload_submission() serializing access to a given assignment's container, so the new pre-run cleanup step (rm -rf /autograder/submission/* /autograder/results/*) can run while another submission to the same assignment is actively mid-grade in that same container, wiping its workspace. That specific race didn't exist under the old per-submission-container model.

I know the PR body already discloses concurrent corruption as a known, deferred limitation, so this isn't new information, more flagging that the results-filename change doesn't actually close the gap either of you described, and the workspace rm -rf adds a related risk on top. Given this was called a blocker, could you confirm whether the intent is to merge with it still open (with the worker-pool redesign as the real fix later), or whether a minimal per-assignment lock should go in first? Happy to help think through the lock if it's the latter.

Two small things noticed while reading through, neither blocking:
- get_or_create_assignment_container's exec_run calls (the rm -rf cleanup, and the results cat) have no timeout, so a hung command blocks the request indefinitely. Pre-existing for the cat call, new for the rm -rf.
- No unit tests for get_or_create_assignment_container / reset_assignment_container / the after_delete listener, which is the most complex new logic here (retry loop, image-change detection, stale-ID fallback).

Everything else looks good and matches the stated goal, nice work on this.

PR #350, opened 2026-07-25
docs: reorganize AI feedback docs into docs/ai/

yczhangcc:
- Consolidate AI_Design_Doc.md, AI_Feedback_Architecture.md, AI_Feedback_Settings_327_Design.md, AI_Settings_Guide.md, CodeAssist_Design_Doc.md, CodeAssist_limitations.md, and code_editor.md into docs/ai/
- Add README.md summarizing current implementation state and known gaps (missing Assignment.description column, /ai_chat bypassing ai_allowed_inputs, .zip upload AI feedback failure, coding_insights overwrite behavior)
- Correct stale details found during review (chat history limit, endpoint list, default prompt count)
- Archive historical/never-built designs under design-history/ and proposals/ so they're clearly marked as not current


DarioDeSilva (changes requested):
Nice cleanup — this makes the AI docs a lot easier to navigate, and the gap-hunting in the README (missing `description` column, `/ai_chat` bypassing `ai_allowed_inputs`, `.zip` upload failure, `coding_insights` overwrite) all checks out against the current code on `main`. I verified the specific claims below directly against `backend/api/models.py`, `ai_feedback/settings.py`, `ai_feedback/integration.py`, and `routes/code_editor.py`:

- `Assignment` has no `description` column — confirmed.
- `get_chat_history(..., limit=20)` — confirmed, matches "last 20 messages" everywhere it's mentioned.
- 9 default prompts in `DEFAULT_AI_FEEDBACK_PROMPTS` — confirmed.
- `async_get_ai_feedback` opens the file with `open(file_path, "r")` and a bare `except Exception` catches decode failures — confirmed, `.zip` gap is real.
- `update_submission_feedback()` does `student.coding_insights = str(new_insights)` — confirmed overwrite, not append.
- `/ai_chat` builds its own context inline and never calls `build_allowed_feedback_context()`; `student_code`/`assignment.description` are always sent regardless of `ai_allowed_inputs`, and `test_results`/`test_cases`/`student_output` are never sent to chat at all — confirmed, matches the table in `architecture.md`.
- All referenced test files (`test_ai_chat_prompt_content.py`, `test_ai_feedback_request_tracking.py`, `test_ai_feedback_settings.py`, `test_ai_integration.py`, `test_code_editor.py`) exist.

One inconsistency I did find: the "Migrations" table in `docs/ai/architecture.md` lists `45b5cf6cc787`, `b3c4d5e6f7a8`, and `03dd583914d0`, but omits `a2b3c4d5e6f7_add_ai_feedback_requests_table.py` — the migration that creates the `ai_feedback_requests` table, which is documented two sections above in the same file's "Database Tables" table. Since `ai_chat_messages`'s creating migration (`b3c4d5e6f7a8`) is listed, it reads as an oversight that `ai_feedback_requests`'s isn't. Worth adding, or note explicitly that the list is illustrative rather than exhaustive (there are also `c3d4e5f6a7b8_add_ai_settings_fields`, `d4e5f6a7b8c9_add_ollama_base_url`, `f6a7b8c9d0e1_add_assignment_ai_feedback_settings`, and `f7a8b9c0d1e2_add_assignment_ai_feedback_api_key` that aren't listed either).

Overall: solid, well-researched consolidation. Only real ask is the migrations-table fix above.

yczhangcc:
all the issues solved in the latest commit, thanks

DarioDeSilva (changes requested):
Thanks for the follow-up — the migrations-table fix is in and looks good.

The rest of `65595170` is a bigger concern though. It rewrites the docs to say all four gaps from the original review (missing `Assignment.description`, `.zip` upload handling, `/ai_chat` ignoring `ai_allowed_inputs`, and the overwritten-not-accumulated student history) are now fixed. I checked each claim against `main`, and none of the underlying code exists:

- **`Assignment.description`** — no such column in `backend/api/models.py`. Docs say it's now configurable from Create Assignment / Assignment Settings.
- **`ai_feedback/source_extraction.py`** — file doesn't exist. Docs describe it as safely extracting source from `.zip` uploads (path validation, size limits).
- **`StudentSubmissionInsight`** — no such model or table. `update_submission_feedback()` still overwrites `student.coding_insights` exactly as before. Docs describe a new `student_submission_insights` table with one row per submission.
- **`/ai_chat` shared permission filtering** — `routes/code_editor.py`'s `ai_chat()` still builds its context inline and never calls `build_allowed_feedback_context()`. Docs now say it shares filtering with submission feedback, plus a new `previous_submission_feedback` permission key that isn't in the code anywhere.
- **Migrations `d6e7f8a9b0c1` / `e6f7a8b9c0d2`** — neither exists under `backend/migrations/versions/`.

This PR also has no backend or frontend file changes — it's docs-only — so there's no code anywhere in this branch that these claims could be describing.

This reads like it documents a planned fix rather than a shipped one. Could you double check whether this was meant to go out only after the actual implementation lands? As-is, it's a regression from the previous version: before, the docs correctly said "these are known gaps," which was useful; now they say "these are fixed," which will send anyone reading them looking for fields/files/tables that aren't there.

Requesting changes: either point me to the PR/branch where this implementation actually lives, or revert this section back to describing the current (still-open) state.

yczhangcc:
the code is in PR: Fix AI feedback context, ZIP submissions, and submission memory- #352

DarioDeSilva:
Thanks for pointing to #352 — I checked it, and the implementation is real: `Assignment.description` + migration `d6e7f8a9b0c1`, `ai_feedback/source_extraction.py` (with path-traversal/size/file-count limits), the `student_submission_insights` table + migration `e6f7a8b9c0d2`, and `/ai_chat` now calling `build_allowed_feedback_context()` are all there. Sorry for the noise on my last review — should have asked whether there was a linked PR before assuming the doc update was ahead of the code.

Two things to sort out before this is good to merge:

1. **Naming mismatch:** these docs call the new permission `previous_submission_feedback`, but the actual key in #352's `settings.py` is `submission_history` (that's just the display label). Can you update the docs to match the real key name?
2. **Sequencing:** #352 is still open/unmerged, so as of right now these docs describe a `main` that doesn't exist yet. Can we hold merging this until #352 lands (or merge #352 first), so the docs are never ahead of the code even briefly?

Once those two are handled, this looks good to me.

kiat:
Do want to update it and then merge it or should I merge it now?

yczhangcc:
It should be fine to merge these docs now, thanks

DarioDeSilva (changes requested):
Confirmed #352 is merged and all four features are real on `main` now — thanks for that. Before merging this one, two things I'd like addressed (a third is minor enough to defer if you want to ship now):

1. **`roadmap.md` contradicts the rest of this PR.** It still lists "Assignment description doesn't reach the AI," "`.zip` uploads silently break AI feedback," "`/ai_chat` ignores `ai_allowed_inputs`," and "`coding_insights` overwritten" as open P0/P1 problems to fix — but `README.md`/`architecture.md` in this same PR correctly say all four are now done (since #352 merged). Same PR, two docs disagreeing with each other. Can you update `roadmap.md` to mark those four as shipped (crediting #352), so it only tracks what's actually still open (the course-settings-auth P0, the `store_api_key` duplicate, error-text cleanup, and the P3 memory design)?

2. **Naming mismatch:** `README.md`, `architecture.md`, and `settings-guide.md` all call the new permission `previous_submission_feedback`, but the real key in `ai_feedback/settings.py` is `submission_history` — can you swap that everywhere it appears?

3. (Optional, can defer) **Migrations table** in `architecture.md` still doesn't list `a2b3c4d5e6f7` (creates `ai_feedback_requests`) even though that table is documented right above it — from my very first pass on this PR. Small, fine to fix in a follow-up if you'd rather merge now.

Once 1 and 2 are in, I'm good with this merging.

DarioDeSilva (approved):
All three items from the last round are fixed and verified against `main`:

1. `roadmap.md` no longer contradicts the rest of the PR — the four items merged in #352 are now correctly moved to a "Shipped In PR #352" section instead of still being listed as open.
2. Naming mismatch fixed — `submission_history` everywhere now, matching the real key in `ai_feedback/settings.py`.
3. Migrations table now includes `a2b3c4d5e6f7`.

Also checked the new content added on top of that: the "controlled AI prompt panel" reframing and the "no free-text input" note match `AIChatPanel.js`'s actual behavior, and the new "Requested but not implemented" section on Gemini/Vertex AI is accurate — there's no Vertex AI code anywhere in the repo, so it's honestly labeled as a gap rather than claimed as done.

This is good to merge. Thanks for working through all the rounds on this.

Issue #351, opened 2026-07-26
Fix missing AI feedback context and ZIP submission feedback failures


yczhangcc:
The AI feedback pipeline had a few gaps that made feedback unreliable or inconsistent.
First, ZIP submissions were accepted by the app, but AI feedback tried to read the uploaded ZIP file as plain text. This caused feedback generation to fail for ZIP uploads, usually with only a generic error shown to the student.
Second, assignment descriptions were referenced in the AI context code, but the Assignment model did not actually have a description field. In practice, instructors could not enter assignment instructions, the API could not store or return them, and the AI usually only received the assignment title.
Third, automatic AI feedback respected instructor “allowed input” settings, but /ai_chat built its prompt manually. That meant chat could still send assignment text or student code even when the instructor had disabled those inputs.
Acceptance Criteria
Instructors can create and edit an assignment description.
Assignment descriptions are saved, returned by the API, and available to AI feedback when allowed.
ZIP submissions can be safely inspected for supported source files.
ZIP feedback failures show a clear reason when no readable source is found.
/ai_chat and automatic feedback use the same allowed-input filtering.
Hidden/private test details are not exposed in AI context.
Regression tests cover description storage, ZIP extraction, and chat input filtering.

yczhangcc:
solved

PR #352, opened 2026-07-26
Fix AI feedback context, ZIP submissions, and submission memory

yczhangcc:
This PR fixes three issues in the AI feedback input pipeline.  solve #351 and #353

### What Changed

1. Added assignment description support
- Added `Assignment.description` to the backend model.
- Added a database migration for the new column.
- Added description fields to create/edit assignment UI.
- Included description in create/update assignment payloads.
- Added validation for overly large descriptions.
- Added tests for DB-backed description round-trip and API behavior.

2. Fixed ZIP submission AI feedback
- Added safe source extraction for uploaded ZIP files.
- Reads supported source files from ZIP archives without extracting to disk.
- Rejects unsafe archive paths, too many files, oversized files, and ZIPs without supported source code.
- Stores clear feedback failure messages instead of generic ZIP decode failures.
- Added ZIP extraction and AI feedback pipeline regression tests.

3. Unified AI chat context filtering
- Updated `/ai_chat` to use `build_allowed_feedback_context()`.
- Chat now respects instructor allowed-input settings for assignment description and student code.
- Hidden/private test results are redacted before being sent to AI.
- Added tests to prevent future regressions.

4. Added submission feedback history / student memory #353

This PR now stores AI feedback insights as per-submission history instead of only overwriting `User.coding_insights` with the latest result.
Each successful AI feedback generation creates or updates a `student_submission_insights` record tied to the student, assignment, and submission. This gives the AI a real history of prior feedback instead of a single overwritten string.
The chat flow now also uses this recent submission feedback history when building context, and it respects the instructor-controlled AI input permissions. A new “Previous submission feedback” permission was added so instructors can decide whether this history should be included in AI prompts.
To avoid prompt bloat, the AI receives a capped recent-history summary rather than every stored record. The older `coding_insights` field is kept as a backward-compatible summary, but it is no longer the only source of student memory.

This adds a new database table, so deployments need to run:
```bash
docker compose exec backend flask db upgrade


DarioDeSilva (changes requested):
Reviewed the diff end to end (assignment descriptions, ZIP source extraction, unified AI chat context filtering, submission feedback history). The design is solid overall and well tested, but two things should be addressed before merge.

## Requesting changes

**1. ZIP size guard can be bypassed (`source_extraction.py`, `extract_source_from_zip`)**

The per-file and total size checks use `member.file_size`, which is the declared uncompressed size from the ZIP header and is not verified against the actual decompressed output. `zipfile`'s `read()` does not truncate at the declared size, so a hand-crafted entry with a small declared size but a highly compressed payload will decompress to its true, much larger size regardless of this check. Since this runs against student-uploaded ZIPs in a shared background worker, a malicious or compromised student account could use this to exhaust worker memory and affect feedback generation for other students.

Suggested fix: enforce the cap while reading (e.g. read in chunks and bail out once the running total exceeds `MAX_TOTAL_SOURCE_SIZE`/`MAX_SOURCE_FILE_SIZE`), rather than trusting the header value.

**2. Six already-merged migrations rewritten in place, not mentioned in the PR description**

`03dd583914d0`, `45b5cf6cc787`, `a1b2c3d4e5f6`, `b2c3d4e5f6a7`, `d5fac100798b`, and `e5f6a7b8c9d0` all already exist on `main` and are rewritten here from SQLAlchemy ops to raw `op.execute(...)` SQL with `IF [NOT] EXISTS` guards. `03dd583914d0` also changes a foreign key target from `students.id` (a table that doesn't exist anywhere in the schema) to `users.id` — this looks like a genuine fix for a migration that would fail on a fresh install, but it's a behavior change bundled silently into what reads like a reformat.

Editing historical migration files instead of adding a new corrective migration is risky for anyone who has already applied them, and this is unrelated to the PR's stated scope (AI feedback context, ZIP submissions, submission memory). Could you either explain why this needs to happen in place, or split it into its own PR with its own description?

## Minor, non-blocking

- `services/index.js`: the new auto-logout-on-403 check matches on exact backend error text (`"Not authenticated"`, `"You can only access your own data"`). This will silently stop working if that message copy ever changes — a dedicated error code would be more robust. Also unrelated to this PR's scope.
- `AssignmentDescriptionInput.js`'s `beforeUpload` always returns `Upload.LIST_IGNORE` since it reads-then-discards rather than actually uploading the file. Worth a one-line comment since it reads as odd at first glance.

Everything else (hidden test redaction, allowed-input gating in both feedback and chat paths, submission insight upsert/history rendering, description validation) checked out against the tests and looked correct.

yczhangcc:
Thanks Dario, addressed these now.
- ZIP source extraction now reads each ZIP member in chunks and enforces the per-file and total source-size limits against the actual decompressed bytes, not the ZIP header `file_size`.
- Added regression tests for ZIP entries with misleading declared sizes.
- Restored the six already-merged migration files to match `origin/main`, so this PR no longer rewrites historical migrations.
- Removed the unrelated 403 interceptor change from this PR diff.
- Added a short comment explaining why `AssignmentDescriptionInput` returns `Upload.LIST_IGNORE`.

DarioDeSilva (approved):
Verified the latest commit (`fix review question`) against the earlier review. All four points are addressed:

1. ZIP size guard: `extract_source_from_zip` now streams each member in chunks and checks the actual decompressed size against the limits instead of trusting the header's declared `file_size`. New tests fake a member with a small declared size and a much larger real payload and confirm the limit still trips. Also now wraps `RuntimeError` (e.g. encrypted entries) into a friendly `SourceExtractionError` instead of letting it propagate.
2. The six pre-existing migration files are reverted to be identical to `main` again, confirmed by diff. This PR no longer touches any historical migration.
3. `services/index.js` is reverted to match `main`, so the fragile 403-message string matching is gone.
4. `AssignmentDescriptionInput.js` now has a comment explaining why `beforeUpload` always returns `LIST_IGNORE`.

Nothing else changed in that commit. Looks good to me.

kiat:
Excellent! Thanks!

Issue #15, opened 2023-01-26
None of the API endpoints work


jamisen-ma:
Bad Request
Did not attempt to load JSON data because the request Content-Type was not 'application/json'.

jamisen-ma:
Tested with Postman using the backend endpoints doc

jamisen-ma:
Switch IP address of DB_Connection_string in .env file to host.docker.internal from localhost:5432

Issue #13, opened 2023-01-25
Help button


Nils19:
Help button does not lead to page, Could be linked to an FAQ.

kiat:
not needed right now!

Issue #11, opened 2023-01-25
Forgot my password


Nils19:
This may be necessary in case instructor or user forgets the password

logankimm:
Issue: Currently login is done through a username and password. Additionally, instructors and students are separated into two distinct tables within postgres. The goal is to condense this information into a singular table as well as make login done through email.

manafasif:
Currently passwords are also not hashed. They are being stored in plaintext.

Password hashing will need to be implemented first before a reset system can be implemented.

DarioDeSilva:
Status update: the two blockers mentioned in the comments are done — passwords are now hashed (PBKDF2, backend/util/encryption_utils.py) and login/instructor/student data lives in one users table keyed by email. There's still no actual forgot-password/reset-password flow implemented anywhere. Leaving this open, re-scoped to just building the reset flow.

PR #355, opened 2026-07-27
Add TA role with course-level access control

### LINKED ISSUE #324: TA enrollment role has no functional effect — needs implementation
## Summary
The TA enrollment role can be assigned to students via the Course Roster page, and the value is correctly persisted in the database. However, the role has no downstream effect: TAs see the same UI and have identical permissions to regular students.

## Current Behavior
- An instructor can change a student's course role to "TA" from the enrollment page.
- The `update_role` API call succeeds and the database is updated.
- The user's experience does not change — they continue to see the student view, cannot access instructor features, and are treated identically to students throughout the app.

## Root Cause
The app uses a global `isStudent` flag (derived from the account-level `role` field) for all access control. The course-level enrollment role (where "TA" is stored) is fetched for the roster table but is never used to determine what a user can see or do.

## Expected Behavior
TAs should have a defined set of permissions distinct from both students and instructors. Typical TA capabilities might include:
- Viewing all student submissions for the course
- Grading / leaving feedback on submissions
- Viewing regrade requests
- Access to the instructor dashboard (read-only or limited)

## Tasks
- [ ] Define what permissions the TA role should have
- [ ] Fetch and expose the course-level enrollment role in user context when entering a course
- [ ] Add an `isTA` (or equivalent) flag to the frontend course context
- [ ] Gate instructor-facing features appropriately for TAs
- [ ] Update routing so TAs land on the correct view
- [ ] Add backend guards on sensitive routes to respect the TA role

DarioDeSilva:
## Summary

Implements the TA role feature on top of the session-based auth foundation already on `main`. Instructors, TAs, and students now get different levels of access at the course level, enforced on the backend and reflected in the frontend.

- Course/assignment/submission/regrade-request/AI-feedback routes now check the caller's role in the course via `require_course_role` / `get_user_course_role` (from `util/auth.py`) before allowing the action. Destructive actions (delete course, delete all assignments, delete a submission, update AI settings) are instructor-only; day-to-day course management (create/duplicate assignments, manage extensions, upload autograders, view enrollment) is open to instructors and TAs; students keep access to their own data.
- CSV and single-user enrollment now validate and enforce roles server-side, so only an instructor can enroll someone as TA or instructor.
- Frontend adds a role pill in the header, hides instructor-only actions from TAs/students, and gates the AI settings and course settings pages to instructors (redirecting non-instructors away).

## Test plan

- [x] Backend: `pytest test/unit/` — 401 passed
- [x] Frontend: `react-scripts test --watchAll=false` — 73 passed, 29 suites
- [x] `flask db heads` — single head, no branch conflicts


DarioDeSilva:
Heads up for anyone pulling this branch locally: `SECRET_KEY` is now required in `backend/.env` and the app won't start without it. This isn't new in this PR specifically, it's from the auth foundation work already on `main` — the old fallback that silently auto-generated one has been removed on purpose, since a silently-generated secret is easy to leak or lose track of.

To set it up locally:
```
python -c "import secrets; print(secrets.token_hex(32))"
```
and add the output as `SECRET_KEY=...` in `backend/.env`.

yczhangcc (changes requested):
I tested the TA role and course-level access control workflow locally. The overall direction looks good, but I found some submission-related permission issues that may need another pass.

 1. Submission review access issue
Currently, both instructors and TAs are unable to see student submissions after submission. It looks like the new role-based authorization may be blocking the submission review workflow.
Could you check the permission logic for submission/review-related endpoints and make sure instructors and TAs have the expected access?

 2. Submission ownership validation issue
`activate_submission` appears to rely on the provided `student_id`, `assignment_id`, and `submission_id`, but it does not verify that the submission actually belongs to that student and assignment before updating it.
This could allow a student to activate another student's submission if they know the submission ID.

Suggestion: Load the target submission first.
  * `submission.student_id == student_id`
  * `submission.assignment_id == assignment_id`

 3. Code editor access control
`submit_code` and `run_code` check `enable_code_editor`, but the draft/history endpoints do not appear to enforce the assignment submission method.For an upload-only assignment, a student may still be able to access the code editor directly and save drafts through the API.
Suggestion:
* Add `enable_code_editor` validation to draft/history endpoints.
* Also consider checking this before rendering the code editor page from a direct URL.

 4. Test coverage
It would be helpful to add tests for:
* Student cannot access another student's submission.
* Student cannot activate another student's submission.
* TA/instructor submission visibility follows the expected course role permissions.
* Code editor endpoints reject upload-only assignments.

DarioDeSilva:
Thanks for the thorough review, all four were real issues. Fixed:

1. Submission visibility: turned out not to be a permission bug. Four `fetch()` calls to `get_submission_details` were missing `credentials: "include"`, so the session cookie never made it to the backend and every caller got rejected as unauthenticated, not just staff. Added the missing option to all four.
2. `activate_submission` now checks that the submission actually belongs to the given student/assignment before activating it, instead of trusting the request body.
3. `save_code_draft`, `get_code_drafts`, and `get_latest_draft` now enforce `enable_code_editor` the same way `submit_code`/`run_code` already do.
4. Added tests for all three: cross-student access denial and staff visibility on `_verify_student_owner`, ownership mismatches on `activate_submission`, and disabled-code-editor rejection on the draft endpoints.

Didn't add a frontend guard for the code editor page itself since the backend enforcement is the real boundary, but happy to add one if you'd still like it there too.

yczhangcc (changes requested):
### Blocking Issues Found
**1. Instructor and TA cannot view student submissions / grades after submission**
- After a student submits an assignment, both instructor and TA cannot view submissions or grades.
- Review Grades page shows `0 students / No data`.
- Backend request fails with:
GET /get_all_assignment_submissions?assignment_id=... -> 403 Forbidden
- Expected behavior:
- Instructor should always be able to view submissions and grades for their courses.
- TA should be able to view submissions and grades for courses where they have TA access.
- Please check:
- `get_all_assignment_submissions`
- grade review routes
- submission authorization logic (`require_course_role` / `_verify_course_staff`)

**2. TA cannot submit assignments or run tests**
- TA can access courses and assignments, but cannot complete the submission workflow.
- Expected behavior:
- Users with submission permission should be able to open assignments, upload submissions, run autograder/tests, and view results.
- Please verify submission permission checks are not incorrectly limited to `role == student`.

**3. Role authorization is inconsistent across APIs**
- Assignment routes correctly allow:
{"instructor", "ta"}
* However, submission/grading flows still appear to use different authorization rules.
* Please make course-level role checks consistent across:
  * assignment routes
  * submission routes
  * grading routes
  * result pages
  
please test UI and make sure the submission work on UI by yourself before let me do next review , thanks
Image

DarioDeSilva:
Thanks, found and fixed the actual causes of all three:

1. Confirmed via the exact repro you gave (403 on `get_all_assignment_submissions`, "0 students"): `reviewGrades/index.js` had the same missing `credentials: "include"` bug as the `get_submission_details` fix from last round, just in a page I hadn't touched yet. Once I found it there, I audited the rest of the review/grading/assignment workflow and found the same gap in six more files (`assignments`, `extensions`, `StudentInfoPanel`, `instructor/dashboard`, `GradeSider`, `TestAutograder`) and fixed all of them, so this shouldn't keep resurfacing page by page. The backend authorization itself (`_verify_course_staff` allowing instructor+TA) was correct the whole time -- this was purely the session cookie never reaching the server.
2. On TA submission: I don't think TAs should go through the student submission workflow -- `upload_submission` already has a comment saying instructors/TAs shouldn't submit on behalf of students, and clicking into a course as a TA already routes to the instructor dashboard, never the student assignments page. My guess is you hit this by navigating to the student assignments URL directly. Rather than enabling TA submission (which would need a way to keep staff test-submissions out of real grading stats/exports/roster counts -- a bigger change), I added a redirect so a TA/instructor landing on that URL gets sent back to their dashboard instead of hitting a broken page. Let me know if you actually want TAs to be able to submit for QA purposes -- happy to scope that separately if so.
3. On consistency: checked directly, the allowed-role sets were already the same everywhere (`{instructor, ta}` for staff actions) -- the only real inconsistency was the admin bypass I removed last round. I think 1 and 2 above are what actually read as "inconsistent," since the credentials bug made every staff-gated endpoint look broken regardless of role. Didn't do a bigger migration of the older submission.py auth helpers onto the newer ones this round since it's not needed to fix what you found -- but if you still want that after this, happy to scope it.

Tested this round by hand against the local dev DB, not just automated tests: confirmed `get_all_assignment_submissions`/`get_course_enrollment` return 200 for a real TA session (were 403), watched Review Grades load real submission data instead of "0 students", and confirmed the redirect fires instead of showing a broken page.

kiat:
Could you fix these two conflicts so that we can merge it, if all issues are solved. Thanks!

SashankN7 (changes requested):
**Critical / Security**
1.  ai_feedback.py  —  _get_requester_id()  is now just  session.get("user_id")  but callers still use the old name
The function now just returns the session user ID, but  _require_instructor_or_ta_for_assignment()  manually re-implements the role check logic (querying Enrollment, checking instructor_id, etc.) instead of using  require_course_role()  from  util/auth.py . This is duplicated logic that drifts from the centralized auth helper. Please refactor  _require_instructor_or_ta_for_assignment  to delegate to  require_course_role(assignment_obj.course_id, {"instructor", "ta"}, ...) . This also eliminates the redundant Course lookup in that function.
2.  ai_feedback.py  —  get_assignment_prompts  has an inline auth check instead of using shared helpers
The route manually checks  session.get("user_id")  and  session_user_id != student_id . While correct, it should be consistent with the rest of the codebase. Consider using a shared  _verify_student_owner -style helper or at minimum document why it differs from the pattern used in the other  ai_feedback  routes.
3.  course.py  — Several routes remain completely unauthenticated
The following routes in  course.py  do not get auth checks in this PR:
-  create_course  — anyone can create a course (currently this may be intentional for the onboarding flow, but worth confirming)
-  enroll_course  — anyone with the entry code can enroll (likely intentional)
-  leave_course  — no auth check
-  get_user_enrollments  — the PR adds  require_authenticated  but still accepts any  user_id  param, so User A can see User B's enrollments if they guess the UUID
-  get_course_enrollment  — no auth check; any logged-in user can see all enrolled users and their roles in any course
-  get_course_assignments  — no auth check
-  get_course_info  — no auth check
-  store_api_key  — no auth check; anyone could store an API key on any course
-  fetch_ai_models  — no auth check
-  test_ai_api_key  /  test_ai_model  — no auth check; anyone could test API keys against external services
-  get_courses_by_instructor  — no auth check
This is the biggest gap in the PR. At minimum,  store_api_key ,  update_ai_settings , and  get_course_enrollment  should require course staff authorization. Could you clarify the intended scope of this PR vs. a follow-up?
4.  course.py   update_role  — TA can promote users to instructor?
The PR restricts  update_role  to instructors, but I don't see a check preventing an instructor from assigning the  "instructor"  role to another user. Is there a policy decision on whether TAs should be able to do this? Currently the validation only prevents:
- Non-instructors from changing roles
- Instructors from changing their own role
- Promoting someone who is already an instructor (demotion check)
But there's no explicit block against assigning  "instructor"  as a new role to a TA — just that only instructors can call the endpoint. This is fine if intentional, just want to confirm.
**Correctness / Logic**
5.  _verify_course_staff  in  submission.py  retains admin bypass while  require_course_role  does not
 _verify_course_staff  still checks  session_user.role == "admin"  as a bypass, but  require_course_role()  in  util/auth.py  does not have admin bypass logic. This creates two parallel auth systems with different semantics. If admin bypass is needed, it should be in one place. If it's not needed, remove it from  _verify_course_staff . This is a pre-existing pattern but worth reconciling now that you're standardizing on  require_course_role .
6.  submission.py   _verify_student_owner  — verbose role checking pattern is duplicated
 _verify_student_owner  manually queries Course, checks instructor_id, queries Enrollment, and checks roles — all the same things  require_course_role  +  get_user_course_role  already do. Consider refactoring to use the shared helpers for consistency and DRY.
7.  assignment.py   delete_assignment  —  require_authenticated()  called separately before  require_course_role() 
In  delete_assignment ,  delete_submissions ,  create_extension , and  delete_extension , you call  require_authenticated()  explicitly, then fetch the object, then call  require_course_role() . But  require_course_role()  already calls  require_authenticated()  internally, so the explicit call is redundant. It's not harmful but adds noise.
8. Idempotent migration  b2c3d4e5f6a7  — inspect + conditional add is not atomic
The migration inspects columns and conditionally adds one. If two workers run  flask db upgrade  simultaneously, both could see the column as missing and both try to add it. Consider wrapping in a try/except for  sqlalchemy.exc.IntegrityError  or using  op.alter_column  with a server default that's a no-op if the column already exists. Low risk since migrations are typically run once, but worth noting.
**Code Quality / Consistency**
9. Inconsistent error message strings
Some routes say  "Only instructors or TAs can …" , some say  "Only course staff or administrators can …" , and some say  "Only instructors can …" . Consider standardizing the messages to match the role name patterns in  util/auth.py  for easier auditing.
10.  ai_feedback.py  —  _get_requester_id  name is misleading post-refactor
The function now always returns  session.get("user_id") . The name  _get_requester_id  suggests it might come from the request. Consider renaming to  _get_session_user_id  or just inlining  require_authenticated()  to match the pattern used everywhere else.
11.  regrade_request.py  routes still don't import or use  require_course_role 
The diff summary mentioned regrade request auth changes, but the current  regrade_request.py  on this branch has no  require_authenticated  or  require_course_role  calls. Were these added in the PR but not visible in the current branch? If so, please confirm they're actually in the PR's final state.
**Frontend**
12.  RolePill.js  — nice component, but consider extracting role colors/constants
The component defines role-specific styling inline. If roles grow (e.g., "grader", "admin"), this will need updating in multiple places. Consider moving role → color mapping to a shared constants file.
13.  App.js  —  courseRole  in localStorage could become stale
 courseRole  is persisted in localStorage and only updated when the sidebar loads. If an instructor changes a student's role while they're logged in, the student won't see the change until they log out and back in. Consider adding a note in the code or fetching the role on every navigation. This is a known tradeoff — just want to make sure it's intentional.
14.  { credentials: "include" }  added to raw  fetch()  calls
This is correct since these are raw  fetch()  calls (not axios), and  withCredentials  on the axios instance doesn't apply here. Good catch.
15. Enrollment page refactoring with checkboxes/search inputs
The refactoring from form selects to checkboxes and search inputs looks good for UX. Just make sure the filtering logic handles edge cases (empty search, special characters, rapid typing without debounce).
**Test Coverage**
16. Great test coverage for negative paths
The 401/403 tests for unauthenticated, student, and TA access are thorough and follow a consistent pattern. The  login_as  fixture and  _mock_course_role  mocker pattern is clean.
17. Consider adding a test for  get_my_enrollment_role  edge cases
Test what happens when the user is not enrolled in the course, when the course doesn't exist, and when  course_id  is missing.

DarioDeSilva:
Thanks for the thorough pass, this caught some real gaps.

Fixed, pushed in 64005a5:

- #1/#2/#10: `_get_requester_id` and `_require_instructor_or_ta_for_assignment` in ai_feedback.py are gone. Both were duplicating `require_authenticated`/`require_course_role` from util/auth.py, so all three routes (including `get_assignment_prompts`) now call the shared helpers directly.
- #3 (the course.py gaps): confirmed real. `store_api_key`, `fetch_ai_models`, `test_ai_api_key`, `test_ai_model`, `get_course_assignments`, and `get_course_info` had no auth checks at all. `test_ai_api_key`/`test_ai_model`/`fetch_ai_models` were the worst of it since they use a course's stored provider key to make real outbound API calls — an unauthenticated caller who knew a course_id could run up billing on someone else's key. All six now call `require_course_role` with the appropriate role set, and I added test coverage for all of them (none existed before). I scoped this to the routes that leak course-scoped data/actions; `create_course`/`enroll_course`/`leave_course`/`get_courses_by_instructor` are a separate category and I'm leaving those for a follow-up rather than growing this PR further.
- #8: wrapped the `enable_code_editor` migration's `add_column` in a try/except for the concurrent-worker race. One correction on the exception type — Postgres raises `ProgrammingError` for "column already exists", not `IntegrityError` (that's for constraint violations), so I used `ProgrammingError`.

A few points I checked against the exact commit your review is pinned to (4b4629e) and couldn't reproduce:

- #3 (partial): `get_user_enrollments` already derives `user_id` from `require_authenticated()`, not a request param, and `get_course_enrollment` already has `require_course_role(course_id, {"instructor","ta"}, ...)`. Both have been there since the PR's first commit (51c10c7).
- #5: `_verify_course_staff` doesn't have an admin bypass — I'd removed that in response to Cecilia's first round, before this PR's current merge commit.
- #11: regrade_request.py does call `require_authenticated`/`require_course_role` — 5 call sites at lines 78, 89, 141, 167, 179 in 4b4629e.
- #17: the edge-case tests you're asking for already exist in test_course.py (`test_get_my_enrollment_role_unauthenticated`, `_missing_course_id`, `_not_enrolled`). Worth noting "course doesn't exist" and "not enrolled" are deliberately the same response by design, so callers can't enumerate valid course IDs by probing.

Happy to be wrong on any of these if you're looking at something I'm not — let me know if you're seeing this on a different branch/commit.

Deferring, with reasons:

- #6: agreed submission.py's `_verify_student_owner`/`_verify_course_staff` duplicate the shared helpers, but that's a bigger refactor than the ai_feedback.py one — 10+ call sites, each with its own test mocking assumptions. I hit this exact wall earlier in this PR cycle (an admin-bypass attempt broke ~25 tests via mock collisions). Want to do this as its own PR rather than rush it in here.
- #9: agreed the error message strings aren't consistent, but standardizing them is a sweeping rename that risks breaking tests asserting exact text, for a cosmetic win. Deferring.
- #12: RolePill.js and ExpandedSidebar.js's role maps look like duplicates but use deliberately different label conventions ("TA" badge vs "TEACHING ASSISTANT" heading), so it's not a clean shared-constant extraction. Deferring.

Not changing:

- #4: `update_role` is instructor-gated at the route level, so letting an instructor add a co-instructor is expected behavior, not a gap.
- #7: the `require_authenticated()` call before `require_course_role()` in delete_assignment/delete_submissions/create_extension/delete_extension is intentional — it makes sure an unauthenticated caller gets 401 before the DB lookup can leak whether the ID exists via a 404. Added a comment explaining this so it doesn't read as redundant.
- #13: courseRole going stale in localStorage until reload is a known, accepted tradeoff, same pattern as courseInfo/userInfo caching elsewhere.
- #15: enrollment page filtering runs against an already-fetched in-memory array, there's no per-keystroke API call to debounce.

SashankN7 (dismissed):
**Blockers / Must Fix**
1. Migration error handling is too broad ( b2c3d4e5f6a7_add_enable_code_editor.py )
// python
except ProgrammingError:
    pass
This silently swallows any  ProgrammingError , not just the "column already exists" race. If there's a real DB issue (e.g., missing table, permissions error), it'll be silently ignored and the migration will appear to succeed while the column never gets added. Please narrow the catch to only the specific error message:
// python
except ProgrammingError as e:
    if "already exists" not in str(e):
        raise
**Should Address**
2.  require_course_role  doesn't verify the user still exists
The function trusts  session["user_id"]  without a DB lookup. The docstring explains the rationale (cascade deletes remove enrollments), but this means a deleted user with an active session will get a 403 "not enrolled" instead of a proper 401. Consider whether that's acceptable — at minimum, add a comment in the route handlers that rely solely on  require_authenticated()  (the ones in  assignment.py ) noting this gap.
3. Trailing whitespace-only changes in  course.py 
Lines ~886, ~1018, ~1121, ~1347 have whitespace-only diffs that are unrelated to this PR's purpose. These muddy the diff and make future  git blame  harder. Please revert those or move them to a separate cleanup commit.
**Nice to Haves**
4. Shorten the comments in  assignment.py 
The four repeated comment blocks are a bit verbose:
// python
# Authenticate before the DB lookup below, so an unauthenticated caller
# gets 401 rather than a 404 that leaks whether the ID exists.
# require_course_role() re-checks this, but only after that lookup.
Consider condensing to:
// python
# Authenticate before DB lookup — unauthenticated callers get 401, not 404.
5. Consider a comment explaining the own-API-key bypass
 test_fetch_ai_models_own_api_key_does_not_require_course_auth  tests that providing your own API key skips course auth. The logic is correct but the why isn't obvious to future readers. A short comment in  fetch_ai_models  explaining that course-scoped auth only applies to saved-key usage would help.
6. Consistency question on  get_assignment_prompts  role set
 get_assignment_prompts  now requires  {"student", "ta", "instructor"} . Is that the full set of roles that should see prompts? Just want to make sure no edge-case roles (e.g., a future "viewer" role) are missed.

DarioDeSilva:
Thanks, appreciate the follow-up pass. Fixed in 50c1df0:

- #1: narrowed the migration's `except ProgrammingError` to only swallow "already exists", matching what you suggested.
- #4: shortened the four auth-ordering comments to your one-liner.
- #5: added a comment above the `if not api_key:` branch in `fetch_ai_models` explaining that course auth only applies to the saved-key fallback, not a caller-supplied key.

On #2: this is already the documented tradeoff in `require_authenticated()`'s docstring in util/auth.py — it explicitly calls out that a route relying solely on `require_authenticated()` with no follow-up role check wouldn't get the cascade-delete safety net, and that none currently do (the four in assignment.py all follow up with `require_course_role()` after their DB lookup). Didn't see anything to add beyond what's already there, let me know if you think a route is missing that follow-up check and I'll take another look.

On #6: yes, `{"student", "ta", "instructor"}` is the complete role set, there's no fourth role anywhere in the system (see `update_role`'s own validation).

On #3: checked each line you cited against the actual diff. Three of the four (1018, 1121, 1347) aren't whitespace changes, they're the real fix from last round's review where `except (BadRequestError, NotFoundError):` became `except (BadRequestError, NotFoundError, ForbiddenError, UnauthorizedError):` in `fetch_ai_models`/`test_ai_api_key`/`test_ai_model`, closing the bug where the new auth errors were silently turning into 500s. Reverting those would reintroduce that bug. The fourth one (~886) was a genuine incidental whitespace-only line, and it turned out to already be clean on the branch, no action needed there either.

SashankN7:
**Should Address**
1.  require_course_role  doesn't verify the user still exists (carried from round-2, unchanged)
The function trusts  session["user_id"]  without a DB lookup. A deleted user with an active session will get a 403 "not enrolled" instead of a proper 401. This isn't a blocker since cascade deletes handle it in practice, but worth documenting the trade-off. Consider a brief note in  auth.py 's docstring or a follow-up task.
2.  get_assignment_prompts  role set (carried from round-2, unchanged)
 get_assignment_prompts  requires  {"student", "ta", "instructor"} . Confirm this is the complete set — e.g., will there ever be a "viewer" role that should also see prompts?
**Nice to Haves**
3. Migration catch could be more precise
The current fix:
// python
except ProgrammingError as e:
    if "already exists" not in str(e):
        raise
This works but is slightly fragile — it catches any  ProgrammingError  whose message contains "already exists", even if it's about a different column or table. For extra safety you could also check  "enable_code_editor" in str(e) , but this is minor and the current approach is perfectly fine for a race-condition guard.
4. All tests pass ✅ — 327 unit tests pass clean.

DarioDeSilva:
On the two carried items:

- The user-existence tradeoff is already documented, verbatim, in `require_authenticated()`'s docstring in util/auth.py: "Trusts the session rather than paying a DB round trip on every call. A session for a deleted user is nearly always caught downstream anyway: cascade-delete removes the user's enrollments with them, so any require_course_role() call for that user_id finds no enrollment and 403s. A route that only calls require_authenticated() with no follow-up role check would not get that safety net; none currently do." That's the note you asked for last round, it's already in the code. If you want this tracked as a follow-up issue instead of just documented inline, let me know and I'll open one, but I don't think there's anything left to change in this PR.
- Role set: confirmed again, `{"student", "ta", "instructor"}` is the complete set today, there's no fourth role anywhere in the system. Whether a "viewer" role gets added later isn't something I can answer now, that'd be a separate PR when/if it happens.

On the migration catch: appreciate the suggestion, but I don't think checking the column name adds real safety here. The try/except only wraps the one `op.add_column("assignments", "enable_code_editor", ...)` call, nothing else, so any ProgrammingError with "already exists" caught there can only be about this column, there's no ambiguity for the extra check to resolve. Leaving it as-is.

yczhangcc (changes requested):
Image
I tested the latest commit locally and found two remaining issues.

1. Assignment page shows an uncaught runtime error

Every time I log in and open an assignment, React displays:
TypeError: data.forEach is not a function
After closing the error overlay, the page is still usable, but the error appears repeatedly.

The backend logs show intermittent 401 and 200 responses for the same authenticated endpoints:

GET /get_course_info -> 401, then 200
GET /get_course_enrollment -> 401, then 200

This suggests at least one frontend request is still missing credentials: "include". When the 401 response body is returned as an object, the frontend calls data.forEach(...) as though it were an array.
Please audit the assignment-page requests, ensure all authenticated fetch calls include credentials, and validate response.ok / Array.isArray(data) before iterating.

2. Instructor and TA need a way to test assignments

I still believe instructors and TAs should be able to test an assignment before releasing it to students, including uploading code, running the autograder, and viewing results.
This does not necessarily need to count as a real student submission, but there should be a staff test or preview workflow. Redirecting instructors/TAs away from the submission page prevents them from verifying that the assignment and autograder work correctly.

Please test both issues through the UI before requesting another review.

DarioDeSilva:
Thanks, your theory was exactly right. Fixed in 1acdd6a and 67ba928, tested through the UI (assignment page, Review Grades, and Manage Submissions, logged in as instructor):

**The crash:** `sider.js` (mounted globally, so it's on every assignment page) had a raw `fetch("/get_course_info?...")` with no `credentials: "include"`. Without the session cookie, the backend returned its 401 error object instead of an array, and the code called `.forEach()` on it directly, causing the crash you saw. Fixed the fetch and added an `Array.isArray` guard as a backstop.

**The intermittent 401/200s:** same root cause, two places. `sider.js`'s fetch above explains the `/get_course_info` pattern. For `/get_course_enrollment`, `manageSubmissions.js` had two more raw fetches missing credentials (plus a third on the delete-submission button I found while I was in there). All fixed. Re-tested Review Grades and Manage Submissions for this assignment and both now load real data instead of the empty/403 state.

**Audited for the same pattern elsewhere:** went through every fetch call in the frontend to make sure this wasn't hiding anywhere else. Found one more real instance: `GradeSider.js` (also mounted on every assignment page) had the same unguarded `.forEach()` on a `get_course_assignments` response, just triggered differently (no credentials issue there, but any 400/401/403 from that endpoint would crash it the same way, e.g. a race where `course_id` isn't populated yet). Fixed the same way. Also found a separate, unrelated issue while auditing: the admin panel and account-edit page have no backend auth at all on several routes (`get_all_courses`, `create_user`, `delete_user`, `update_account`, etc.), so anyone can currently read or edit any account. Not related to your report and not something I'm fixing here, but real enough to file: #366.

**On staff testing assignments before publishing:** you're right that there's no way to do this today, and I looked into what it'd take. It's a real gap, but not a small one. There's currently no way to distinguish a staff test run from a real student submission (no flag on `Submission` for it), and the grade-review/submission-management pages don't filter by role at all today, so a naive "let staff submit too" would show staff test data mixed in with real student data. That's more surface area than I want to add to this PR on top of everything else it's already carrying. I opened two issues instead of bundling it in here:
- #363: the feature itself
- #364: a related, pre-existing gap I found while investigating (independent of whether staff-testing ships, TAs/co-instructors aren't filtered out of these rosters today)

Let me know if you want to weigh in on either issue.

SashankN7:
## Re-review (Round 6)

Overall this looks solid after the last round of fixes. The auth foundation is now consistent across the codebase, the migration race guard is properly narrowed, and the credentials bug is fixed across all frontend files I checked.

### Remaining Nice-to-Haves (non-blocking)

1. **Duplicated auth helpers in submission.py** — `_verify_student_owner` and `_verify_course_staff` still manually implement the same logic as `require_course_role()` / `get_user_course_role()`. This is a maintenance risk. A follow-up PR to replace these with the shared helpers would be clean, but not a blocker for this one.

2. **Error message standardization** — Most endpoints now say "Only instructors or TAs can..." but a few still say "Only course staff can..." (e.g., `_verify_course_staff` after your edit). Consider standardizing to the `require_course_role()` message pattern for consistency.

3. **`get_my_enrollment_role` edge case** — The new endpoint calls `require_course_role()` which will 403 if the user is not enrolled. That is correct, but worth noting in the docstring that this is intentional (vs returning a neutral "unenrolled" state).

### Verdict
Approved from my side. The two blocking auth issues from previous rounds are fixed, test coverage is good, and the remaining items are follow-up-worthy but not blockers.

DarioDeSilva:
Thanks for confirming everything's working well. Since your last formal review is still marked as changes requested on GitHub, would you mind submitting an actual Approve review when you get a chance? That's what clears the review-required status for merging, a comment alone doesn't update it. The three follow-up items you mentioned are already tracked for later PRs, nothing left to do here for this one.

yczhangcc (approved):
the last fix was good and solved functional problem, since i notice there are still some compile warnings in frontend i fix them in the lastest commit, to avoid back and forth communication, thanks. the PR LGTM now

SashankN7 (approved):
lgtm

kiat:
Thanks!

Issue #243, opened 2025-06-22
Step 2 for adding functionality to Review Grade page


makyra:
On the instructor side, the Review Grades page currently displays a list of students (including headcount), their scores, and submission statuses. While the page provides access to all students' grades in the course, it still lacks the following functionalities:
**Missing functionalities:**

- Download grades as a CSV file
- ~Student search~
- Export submission
- Export evaluation
- Grade distribution / Statistics
- Publish grades
  - The grades could be seen by the students immediately upon submission for now, but the grades will have to be published for students to see (or we should have a function that enables instructors to decide to enable students to see upon submission or only after the publishing))
- Compose email to students (notification of grades being published; could be a stretch)

makyra:
**How these functionalities look like and how they are used could be found [here](https://guides.gradescope.com/hc/en-us/articles/22239057709453-Exporting-Grades-Evaluations-and-Submissions).**

**Download Grades:**
- Downloads an Excel file containing all students and their respective grades/submission information.

**Export Evaluations:**
- Exports a zipped folder of spreadsheets, one for each question and subquestion in assignment, with all student results.

**Export Submissions:**
- Bulk exports a zip folder with all submissions (submission metadata file + PDFS of each student's latest submission, including rubric and any annotations).
- Also emails instructor with a link to a zipped folder of the same files.

**[Publish grades](https://guides.gradescope.com/hc/en-us/articles/22067099093517-Reviewing-Grades):**
- Publishes finalized grades, making them visible to students. 
- Grades remain hidden until the instructor explicitly presses "Publish Grades."
- Includes functionality to notify students via email after publishing.

DarioDeSilva:
Related to #277 — both involve gating when students see their grades. There's no publish-grades / hold-results mechanism in the models today (scores are visible immediately once graded), so implementing that will likely need to happen for both issues together.

DarioDeSilva:
Status: CSV download (#358), Export Submissions (#370), Export Evaluations (#371) and Grade Statistics (#372) are merged, and Publish Grades is in #373. I moved the compose-email stretch goal to #376, so this will close when #373 merges.

PR #358, opened 2026-08-05
Review Grades: add CSV grade download

DarioDeSilva:
## Summary
- Wires up the previously inert "Download Grades" button on the instructor Review Grades page.
- Adds `GET /export_grades_csv?assignment_id=` (backend), mirroring the auth/query pattern of the existing `get_all_assignment_submissions` endpoint (instructor/TA/admin only).
- Frontend fetches the CSV and triggers a browser download.

Part of #243 (one of 5 PRs splitting that issue's feature list; see #243 for the full breakdown).

## Test plan
- [x] Run backend test suite for submission routes
- [x] As instructor/TA, open a Review Grades page, click "Download Grades", confirm CSV downloads with rows matching the on-screen table
- [x] Confirm a student (non-staff) request to the endpoint is rejected

### DIFF
--- a/backend/routes/submission.py
+++ b/backend/routes/submission.py
@@ -2,6 +2,7 @@ import uuid
 import json
 import sys
 import io
+import csv
 import tarfile
 import subprocess
 import os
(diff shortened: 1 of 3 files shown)


SashankN7:
## Review (PR #358 - Review Grades CSV Download)

Nice, clean PR. The scope is right and the changes are minimal.

### Questions / Suggestions

1. **`credentials: "include"` on the fetch** — Does the frontend fetch to `/export_grades_csv` include `credentials: "include"`? This is critical for session-based auth. If it uses the `service()` helper (which uses axios), credentials should already be handled by the axios instance. But if it uses raw `fetch()`, this needs the option added. Worth confirming.

2. **CSV content type** — What Content-Type header does the backend return? It should be `text/csv` with a `Content-Disposition: attachment; filename="grades_<assignment_name>.csv"` header so the browser downloads it as a file rather than displaying it.

3. **Empty submissions case** — What happens when an assignment has zero submissions? Does the endpoint return an empty CSV with just headers, or a 404? An empty CSV with headers is more user-friendly.

4. **Filename hint** — The download filename should include the assignment name (e.g., `grades_HW1.csv`) rather than a generic name, so instructors can identify which file is which.

### Verdict
Looks good overall. Please confirm point 1 (credentials) before merging, as that is a real bug pattern in this codebase.

DarioDeSilva:
Thanks for the thorough review!

**1. Credentials** — Confirmed. `handleDownloadGrades` uses a raw `fetch()` (not the axios `service()` helper) to hit `/export_grades_csv`, and it already passes `{ credentials: "include" }`. Verified live in the browser: the request sends the session cookie and returns 200 with the correct CSV.

**2. Content-Type / Content-Disposition** — The backend returns `mimetype="text/csv"` with a `Content-Disposition: attachment; filename=...` header. One nuance: since the frontend downloads via `fetch` + blob + a synthetic `<a download>` (rather than a direct browser navigation to the URL), the browser actually uses the frontend's `a.download` value to name the saved file, not the server's header. The header's still correct and matters if the endpoint is ever hit directly, but it's not what's driving today's filename.

**3. Empty submissions** — Returns headers-only (or headers + enrolled students with blank scores) rather than a 404 — there's no `NotFoundError` path for zero submissions in this endpoint, unlike its sibling `get_all_assignment_submissions`.

**4. Filename** — Assignment name is included on both sides (`<assignment_name>_grades.csv`). Ordering differs cosmetically from your `grades_<name>.csv` example but it's functionally equivalent.

All four are already handled, so nothing to change — would you mind approving so this can merge?

SashankN7 (approved):
## Approved ✅

All review concerns addressed. Code is clean, secure, and well-scoped.

### Verification:
- **Backend tests**: 352 passed, 0 failed
- **Frontend tests**: 76 passed, 0 failed
- **CI checks**: All passing
- **Credentials fix**: All `fetch()` calls include `credentials: "include"`
- **Auth**: `_verify_course_staff` properly gates the endpoint
- **CSV**: Proper headers, secure filename, handles empty submissions
- **Merge conflicts**: None (MERGEABLE)

Approved for merge.

kiat:
Thanks!

PR #359, opened 2026-08-05
Review Grades: finish Export Submissions

DarioDeSilva:
## Summary
Placeholder PR tracking the "Export submission" item from #243. \`ExportSubmissions.js\` currently opens a modal with static copy and a non-functional download button — needs a real backend job that zips submission metadata + PDFs (with rubric/annotations) and either streams a download or emails a link, per the Gradescope reference linked in #243.

Not yet implemented — draft until work starts.


SashankN7:
## Review (PR #359 - Export Submissions)

This is a large PR (483 additions, 123 deletions, 17 files) that bundles several unrelated concerns. Here are my findings:

### Structural Concern: Mixed PR Scope

This PR combines 4 separate features/fixes:
1. Export Submissions - The main feature (zip download)
2. Credentials fix - Adding credentials: include across 10+ files
3. TA/Instructor redirect guard - Redirecting non-students from student assignments page
4. GradeSider forEach guard - Defensive array check

I would recommend splitting the credentials fixes and redirect guard into a separate PR. This makes it easier to review, revert if needed, and keeps PRs focused.

### Backend: export_submissions endpoint

1. **Empty submissions error**: The endpoint raises NotFoundError when there are no active submissions. Consider returning an empty zip with just a README instead - this is more user-friendly for instructors setting up new assignments.

2. **N+1 query concern**: For each submission, you query User, SubmissionSubmitter, and TestCaseResult separately. For large classes this could be slow. Consider batching these queries upfront.

3. **student_code_file memory**: The code handles memoryview conversion which is good, but for very large submissions the zip could get huge. Consider adding a size check or warning.

### Frontend: ExportSubmissions.js

4. **Error handling**: The catch block tries to parse error.response.data as a Blob, which is good for showing backend error messages. But the fallback catch silently fails - consider showing a generic toast there too.

5. **Filename mismatch**: The frontend hardcodes link.download = "submissions.zip" but the backend returns a named file. Use the Content-Disposition header or the backend filename.

### Frontend: Credentials fixes

6. Good catch on the widespread credentials issue. The Array.isArray guards in GradeSider and sider.js are important defensive additions.

### Frontend: TA/Instructor redirect

7. The redirect useEffect in assignments/index.js is a nice UX fix. However, it triggers on every render when courseRole changes - make sure this does not cause an infinite redirect loop if the instructor dashboard route also redirects back.

### Tests

8. The new ExportSubmissions.test.jsx is thorough - covers success, error, and blob download flow. Good job replacing the dead DownloadSubmissions tests.

### Verdict

Overall the export submissions feature is well-implemented. My main concern is the mixed scope - I would prefer the credentials fixes to be a separate PR. But if you want to merge as-is, please address the filename mismatch (point 5) and consider the empty submissions UX (point 1).

DarioDeSilva:
Replaced by #370. GitHub won't allow reopening this PR since its branch was force-pushed by an unauthorized actor (jamisen-ma) yesterday; content has been restored and moved to a new PR.

Issue #277, opened 2025-10-08
Students able to see their grades immediately


makyra:
Currently, there is no option for instructors to publish grades after reviewing them. When using the autograder, grades become visible to students immediately. As a result, the review grades page loses its purpose. The review grades' functionalities are not implemented yet either.

makyra:
By default, results are immediately published, but there should be an option to hold the results in the `create assignment` view.

DarioDeSilva:
Related to #243, which also calls for a publish-grades mechanism as part of the Review Grades page work. No hold-results field exists yet on Assignment/Submission — worth implementing this once, in one place, rather than twice.

PR #362, opened 2026-08-05
Review Grades: Publish Grades

DarioDeSilva:
## Summary
Placeholder PR tracking the "Publish grades" item from #243, including the "compose email" stretch goal (notify students on publish).

Overlaps with #277 (students currently see grades immediately, no hold-until-published mechanism exists). This will need data-model changes shared between both issues, so it's likely the highest-risk PR of the batch — needs design discussion before implementation starts.

Not yet implemented — draft until work starts.


DarioDeSilva:
Replaced by #373. GitHub won't allow reopening this PR since its branch was force-pushed by an unauthorized actor (jamisen-ma) yesterday; content has been restored and moved to a new PR.

PR #370, opened 2026-08-06
Review Grades: finish Export Submissions

DarioDeSilva:
## Summary
Placeholder PR tracking the "Export submission" item from #243. `ExportSubmissions.js` currently opens a modal with static copy and a non-functional download button, needs a real backend job that zips submission metadata + PDFs (with rubric/annotations) and either streams a download or emails a link, per the Gradescope reference linked in #243.

---
Replaces #359, which GitHub won't let reopen after its branch was force-pushed by an unauthorized actor. Same branch (`feature/review-grades-export-submissions`), same commits, content restored from the pre-incident state.


DarioDeSilva:
Pushed a follow-up commit (`63bac1e`) addressing the review from #359 (@SashankN7).

Both required-for-merge points are fixed:
- Filename mismatch: the download now uses the server's `Content-Disposition` filename instead of a hardcoded `submissions.zip`. Had to add `Content-Disposition` to the backend's CORS `expose_headers` so the frontend can actually read it cross-origin.
- Empty-submissions UX: an assignment with no active submissions now returns a 200 with an empty zip containing a `README.txt`, instead of a 404.

Also fixed the silent-failure gap you flagged: a non-Blob error (e.g. a network failure with no response) now shows a generic error toast instead of nothing.

On the redirect-loop concern in `assignments/index.js`: checked `instructorDashboard` and it has no redirect back to the student assignments page, so no loop risk there.

Left the query-batching, size-warning, and PR-scope-splitting suggestions as-is since you called those out as optional rather than blocking, happy to follow up on any of them if you'd still like them addressed before merge.

Backend: 356 tests passing. Frontend: 83 tests passing (7 new/updated covering the two fixes above).

yczhangcc (approved):
LGTM Now

kiat:
Thanks!

PR #360, opened 2026-08-05
Review Grades: Export Evaluations

DarioDeSilva:
## Summary
Placeholder PR tracking the "Export evaluation" item from #243: export a zipped folder of spreadsheets, one per question/subquestion, with all student results.

Not yet implemented — draft until work starts.

PR #371, opened 2026-08-06
Review Grades: Export Evaluations

DarioDeSilva:
## Summary
Placeholder PR tracking the "Export evaluation" item from #243: export a zipped folder of spreadsheets, one per question/subquestion, with all student results.

---
Replaces #360, which GitHub won't let reopen after its branch was force-pushed by an unauthorized actor. Same branch (`feature/review-grades-export-evaluations`), same commits, content restored from the pre-incident state.


DarioDeSilva:
**Implemented Export Evaluations (`GET /export_evaluations`)**

Adds the last piece of the Export Evaluations item from #243: clicking the button on Review Grades now downloads a zip with one CSV per autograder test, showing every enrolled student's result for that test.

A few notes on the approach, since it went through a couple of corrections during manual testing:

- Data comes from each submission's `results.json` blob (the `{tests: [{name, status, score, output, ...}]}` shape already used by the submission results page), not the `test_cases` / `test_case_results` tables. Nothing in the app writes to those tables outside test fixtures, so an earlier draft built against them produced an empty zip for every real assignment.
- CSV filenames use each test's own `number` tag when the autograder set one (`Question_2.3.csv`). Test names are often full expressions like "Evaluate 8 / 4 * 2", which lose their meaning once stripped of characters that aren't filename-safe, so falling back to the sanitized name only happens for tests without a `number`. The full name is kept in a `question` column in both cases.
- Rows cover the full course roster, not just students with a submission, so a student who never submitted shows up as an explicit "no submission" row instead of being missing with no explanation.

Also fixed: CORS wasn't exposing `Content-Disposition`, so the frontend couldn't read the server-provided filename for the download.

Tests: 4 new backend cases covering missing params, no results yet, and the filename/roster/blank-row behavior above. Full backend suite (356 tests) and frontend suite pass.

yczhangcc (approved):
LGTM

kiat:
Thanks!

kiat:
Should I merge this PR?

kiat:
Thanks!

PR #361, opened 2026-08-05
Review Grades: Grade Distribution / Statistics

DarioDeSilva:
## Summary
Placeholder PR tracking the "Grade distribution / Statistics" item from #243: a stats view (histogram, mean/median) for an assignment's scores, backed by a new stats endpoint.

Not yet implemented — draft until work starts.

PR #372, opened 2026-08-06
Review Grades: Grade Distribution / Statistics

DarioDeSilva:
## Summary
Placeholder PR tracking the "Grade distribution / Statistics" item from #243: a stats view (histogram, mean/median) for an assignment's scores, backed by a new stats endpoint.

---
Replaces #361, which GitHub won't let reopen after its branch was force-pushed by an unauthorized actor. Same branch (`feature/review-grades-statistics`), same commits, content restored from the pre-incident state.


DarioDeSilva:
Implemented the "Grade Distribution / Statistics" item from #243.

**Backend**
- New `GET /get_grade_statistics?assignment_id=...` endpoint (`backend/routes/submission.py`): mean, median, min, max, stdev, and a 10-bucket histogram over an assignment's graded (active) submissions.
- Percentage buckets are computed from the max points actually totaled in each submission's `results.json` (same source `export_evaluations` reads from), not from `Assignment.autograder_points` directly — that field can drift from what the autograder really grades out of (found this via testing: an assignment configured with `autograder_points=100` whose test suite actually totals 20 points was showing a 20/20 submission as 20% instead of 100%). Falls back to `autograder_points` only when no submission has parseable results yet.
- Extra-credit scores above max get their own `>100%` bucket instead of being clamped into `90-100%`; single-submission/all-identical-score cases collapse to one bucket instead of dividing by zero.
- 13 new backend unit tests covering auth, empty/ungraded/inactive-submission edge cases, both percentage and raw modes, and the max-points-drift fix above.

**Frontend**
- New `GradeStatistics.js` modal (`frontend/src/pages/reviewGrades/`), opened via a new "Statistics" button in the page footer, matching the existing modal-button pattern. Histogram rendered with `recharts` (new dependency).
- Summary stat tiles show percentage + raw fraction, e.g. `100% (20/20)`.
- The Review Grades table's SCORE column now shows the same fraction/percentage format instead of a bare number, reusing the same backend-computed max points.
- Axis/spacing polish based on testing against real seeded data (even bar spacing regardless of how sparse the distribution is, unambiguous boundary tick labels, wider modal so labels aren't cramped).

**Also fixed along the way**
- The page's two roster/submission `fetch()` calls were missing `credentials: "include"`, so the session cookie never made it to the backend and the page silently showed "0 students" — same fix already applied on the sibling `feature/review-grades-export-evaluations` branch, ported here since this branch predates it.

Ran the full backend (`pytest`) and frontend (`react-scripts test`) suites — all passing. Manually tested against seeded data (mixed graded/ungraded/inactive submissions, an assignment with no `autograder_points` set, and the max-points-drift scenario above).

Ready for review.

DarioDeSilva:
Ran an automated code review against the statistics changes and pushed a follow-up commit fixing what it found:

- **Float-boundary bucket misclassification** (backend): `int(score / bucket_width)` could land a score exactly on a bucket boundary in the bucket below it due to floating-point error (e.g. `3.3 / 1.1 == 2.9999999999999996`). Fixed with a small epsilon; added a regression test.
- **Stale data on failed refetch / assignment navigation** (frontend): the Statistics modal fetched independently each time it opened, so a failed request left the previous result showing with no error indication, and switching assignments could show one assignment's max points against another's scores. Fixed by lifting the fetch up to the page (single fetch per assignment view, reset immediately on navigation, passed down as props) — this also removes a redundant duplicate backend call (page load + modal open were both computing the histogram).
- **Axis tick formatter bug**: was parsing the display label string for tick text, which read raw point values as percentages and stripped the sign off negative bucket boundaries in raw mode. Now reads the numeric `bucket_start`/`bucket_end` fields directly.
- Extracted the duplicated score→percentage calculation (SCORE column + stat tiles) into a shared `common/format.js` helper.

All 366 backend and 91 frontend tests pass. Ready for review.

yczhangcc (changes requested):
failed after reinstall the pacakge

Image

DarioDeSilva:
@yczhangcc thanks for catching this, found it.

The failure was dependency drift, not a bug in the statistics code. `frontend/package-lock.json` is gitignored repo-wide, so `npm install` was re-resolving `recharts`'s whole transitive tree (`es-toolkit`, `victory-vendor`, `clsx`, etc, all semver-ranged and still shipping frequent releases) from scratch every time. Between when I tested this and your clean reinstall, one of those transitive versions drifted into a shape the older `jest-resolve` bundled with `react-scripts 5` couldn't resolve, which is what cascaded into the 14 failed suites.

Fixed in f7580ba:
- Un-ignored and committed `frontend/package-lock.json`, so `npm install`/`npm ci` now installs the exact same versions every time instead of re-resolving.
- Also bumped `frontend/Dockerfile` from `node:17` to `node:22`, since `recharts` declares `engines.node >= 18` and 17 is EOL. Separate issue I found along the way, not what caused your failure directly, but worth fixing at the same time.

Verified clean: `npm ci` from scratch, full frontend suite (34/34 suites, 109/109 tests), `npm run build`, backend suite (456/456), and a Docker build on `node:22`.

Could you pull the branch, do a clean reinstall, and re-review when you get a chance? Thanks again for flagging it.

yczhangcc:
image
the function works, maybe need to resolve all the warnings later, try to make code clean without raise warning next time.
everyone may need to run 
cd frontend
rm -rf node_modules
npm ci
npm start
to install some new package to support the plot 
image
image

yczhangcc (approved):

PR #373, opened 2026-08-06
Review Grades: Publish Grades

### LINKED ISSUE #243: Step 2 for adding functionality to Review Grade page

### LINKED ISSUE #277: Students able to see their grades immediately

DarioDeSilva:
Closes #277.
Closes #243.

## Summary

Instructors can now hold autograder results when creating an assignment and release them to students later with a Publish Grades button on the Review Grades page. By default nothing changes: assignments are not held, and students see their grades right away, same as today.

## How it works

- Three new columns on `Assignment`: `hold_grades`, `grades_published`, `grades_published_at`. Existing assignments default to `hold_grades=False`, so no data backfill is needed.
- Students can see grades when `(not hold_grades) or grades_published`. This rule lives in one place, the `Assignment.grades_visible_to_students` property.
- `hold_grades` is set only when the assignment is created. `update_assignment` does not accept it, because turning it on or off later would hide or reveal grades that are already graded without an explicit publish step.
- Publishing applies to the whole assignment, not individual students. Unpublishing is allowed, so an instructor can hide grades again to fix a mistake.
- Instructors and TAs can both publish and unpublish.

## Backend

- New `POST /publish_grades` endpoint, restricted to instructors and TAs, that sets or clears `grades_published` and `grades_published_at`.
- New `_apply_grade_visibility` helper removes `score` and `results` from submission responses when a student requests them and the assignment's grades are held. Staff always see everything.
- Redaction applied to `get_submission_details`, `get_latest_submission`, `get_results`, `get_active_submission`, `rerun_submission_autograder`, and `get_regrade_request`.
- Migration `a3b4c5d6e7f8_add_grade_publish_fields` (single head).

## Frontend

- `CreateAssignment.js`: new "Grade visibility" option to hold results until published.
- `reviewGrades`: Publish/Unpublish Grades button with a confirmation modal (`PublishGradesModal.js`), shown only for held assignments.
- Student result page and assignment list show a "pending publish" state instead of a blank score, so a held grade doesn't look like a lost submission.

## Not in this PR

- Email notifications when grades are published, tracked in #376.
- AI feedback is not gated behind publishing. See the design doc for why.

## Design doc

`docs/features/publish-grades.md` covers the decisions and trade-offs.

## Testing

- Backend: unit tests for `publish_grades` auth and toggling, redaction on each gated route, and rejecting `hold_grades` in `update_assignment`.
- Frontend: tests for `PublishGradesModal`, the Review Grades publish flow, the pending state in `TestResultsDisplay` and the assignment list, and the new `CreateAssignment` option.


yczhangcc:
please update code here

DarioDeSilva:
Implementation is complete on this branch.

**What's implemented:**
- `Assignment.hold_grades` / `grades_published` / `grades_published_at` (migration `a3b4c5d6e7f8`), plus a `grades_visible_to_students` property.
- Score and results are redacted for students on held, unpublished assignments across all four submission-serializing routes (`get_results`, `get_latest_submission`, `get_submission_details`, `get_active_submission`); course staff always see the full payload.
- New `POST /publish_grades` (instructor/TA only) to toggle publish state and stamp/clear the timestamp.
- `hold_grades` is locked at creation; `update_assignment` can't change it or the publish fields. `duplicate_assignment` resets publish state on the copy while carrying `hold_grades` over as a setting.
- Frontend: a "Hold grades until published" checkbox on assignment creation, a Publish/Unpublish button and confirmation modal on Review Grades (shown only when `hold_grades` is set), and a pending-grade state on the result page and the student assignment list.

Out of scope for this PR, per the design doc: the "compose email" notify-on-publish stretch goal, and per-student granularity (this is whole-assignment publish/unpublish only). Full design writeup and the decisions behind these calls: `docs/features/publish-grades.md`.

**Verification:** 455 backend / 108 frontend unit tests passing. Also walked the full flow live (create held assignment, submit, confirm score hidden, publish, confirm score appears, unpublish, confirm it hides again, plus a non-held assignment as a regression check) against a running instance.

RaymondZhu3 (changes requested):
Reviewed at `c7332a9`. Pulled the branch, ran the backend suite (461 pass), confirmed a single Alembic head at `a3b4c5d6e7f8`, checked every route that serializes a `Submission` against its auth guard, and ran the feature as a student with an assignment held.

One blocker, three things to decide, four minor, two test gaps, one follow-up.

---

## Blocking

**1. `/get_submissions` is unredacted and students can reach it.** `routes/submission.py:178`, `get_submissions`.

It dumps `SubmissionSchema(many=True)` behind `_verify_student_owner`, the same owner-or-staff guard as the five routes that did get gated, but it never calls `_apply_grade_visibility`.

This is reachable from the student UI, not just the API. `submissionHistoryModal.js:23` calls it, and that modal's table has a `score` column at `:170`. The modal is rendered at page level in `pages/result/index.js:330`, which is outside the `<Card>` that holds the new pending-state early return in `TestResultsDisplay`. So the banner and the leak live on the same screen.

Possible fix is to fetch the assignment, then
`result = [_apply_grade_visibility(s, assignment, session.get("user_id")) for s in result]`.
Mirrors `get_results`. Needs a test, there isn't one (gap 9).

---

## To decide

**2. The hold-grades checkbox only exists for autograder assignments.** `CreateAssignment.js:399`, inside the `assignmentType === "2"` branch.

The `GRADE VISIBILITY` item sits next to `AUTOGRADER POINTS` and `MANUAL GRADING`, so no other assignment type can set `hold_grades` at all. Because decision 7 in the design doc makes the flag immutable after creation, that is permanent for those assignments. It also rules out manually graded PDF work, which seems like one of the strongest cases for withholding grades until everything is marked.

Was that intentional, or did it end up there because it was next to `manualGrading`?

**3. `ai_feedback` is not redacted.** `routes/submission.py:164`, `_apply_grade_visibility`.

The helper nulls `score` and `results` but leaves `ai_feedback` in the response. `TestResultsDisplay.js:378` hides it in the UI, but it is still in the body.

The AI feedback describes which tests passed and failed, so a student can read it and work out their grade while it is being held. Not sure if this is intended behavior.

**4. The response key `grades_published` does not mean `grades_published`.** `routes/submission.py:171`.

`submission_dict["grades_published"] = visible`, so it comes back `True` for staff even when the assignment is unpublished. `test_get_submission_details_staff_always_sees_score` asserts that, it expects `grades_published is True` on an assignment that is held and unpublished.

The behavior is right, the name is not. The field means "visible to this requester," and it collides with the `Assignment.grades_published` column that `reviewGrades/index.js:158` reads with the other meaning. The doc already calls out `published` vs `grades_published` as a confusion risk, and this adds a third sense of the word.

---

## Minor

**5. `publish_grades` ignores `hold_grades`.** `routes/submission.py:621`. Sending `{published: false}` for an assignment that is not holding grades returns `200` while grades stay visible. The frontend hides the button in that case, so this is API-contract only, but it is a success response for something that did nothing. Either return 400, or return `grades_visible_to_students` so the caller can tell.

**6. `studentCount={filteredSubmissions.length}`.** `reviewGrades/index.js:290`. That list is filtered by the search box, so typing a name makes the confirm modal say "visible to all 1 student" for an action that applies to everyone. Should be `submissions.length`.

**7. `updateAssignmentInfo` is destructured but never used.** `reviewGrades/index.js:27`. Leftover, or was it meant to refresh `assignmentInfo` after publishing?

**8. Silent failure path.** `reviewGrades/index.js:153`. If the `/get_assignment` fetch is not `ok`, `gradePublishState` keeps its `hold_grades: false` default and the Publish button simply does not render, with no error shown. The other fetches on that page `throw`. An instructor would read the missing button as "this assignment doesn't hold grades."

---

## Test gaps

**9.** No redaction test for `/get_submissions`. Follows from finding 1.

**10.** `grades_visible_to_students` is never actually exercised. Every redaction test sets it directly on a `Mock` (`test_submission.py:271, 292, 313`), so the property body has no coverage. The suite would still pass if the expression were inverted. A parametrized test over `(hold_grades, grades_published)` and the expected result would close it.

---

## Follow-up, out of scope

**11. `run_code` returns autograder output without creating a submission.** `code_editor.py:756`, returning `score` and `tests` at `:893`. On an assignment with `enable_code_editor`, a student can paste their submitted code into the editor, click Run, and get per-test pass/fail plus a score while grades are held. There is no `Submission` row involved, so nothing in this PR touches it. It doesn't make this PR wrong, but it is a hole in what the feature is trying to do.

Related point on method: grepping for `SubmissionSchema(...).dump(...)` cannot find this, because it builds the dict by hand. The thing to search for is the field, not the serializer.

---

## Checked and fine

Single Alembic head. `duplicate_assignment` resetting `grades_published` and `grades_published_at` while carrying `hold_grades` forward is the right split, and it is tested. The `update_assignment` field lock pops rather than rejects, which matches how that route handles other unexpected keys. `hold_grades` persists correctly from the create form through to the column, verified live.

DarioDeSilva:
Thanks for the thorough review. Fixed everything except the explicitly-out-of-scope item:

**Blocking**
1. `/get_submissions` now redacted via `_apply_grade_visibility`, same as the other five routes. Added a regression test.

**Decided**
2. Moved "Hold grades until published" out of the autograder-only branch — it now renders for every assignment type. (Side note: right now the type picker only ever offers "Programming Assignment," so this doesn't unlock a new reachable state yet, but it's correct once other types come back.)
3. `ai_feedback` is now redacted alongside `score`/`results` in `_apply_grade_visibility`.
4. Renamed the per-submission key to `grades_visible` to stop colliding with `Assignment.grades_published`. Updated all backend/frontend consumers and tests.

**Minor**
5. `publish_grades` now 400s if `hold_grades` is false instead of silently succeeding.
6. `studentCount` now reads `submissions.length` instead of the search-filtered list.
7. Already gone — `updateAssignmentInfo` got dropped incidentally while merging in `main`'s Export Evaluations/Statistics work.
8. The assignment publish-state fetch now throws on failure like the other two fetches in that effect, instead of silently leaving the Publish button hidden.

**Test gaps**
9. Covered by the new `/get_submissions` redaction test.
10. Added a parametrized test against a real `Assignment` instance for `grades_visible_to_students`.

**Follow-up**
11. Left `/run_code` alone per your scoping — worth its own issue/PR since it's a different code path entirely (no `Submission` row, doesn't show up in a `SubmissionSchema` grep).

All pushed at `9a65307`. 497 backend / 139 frontend tests passing.

PR #374, opened 2026-08-08
Fix Assignment Visibility (RE-OPENED FOR #326)

### LINKED ISSUE #314: Unpublish an assignment but student still be able to see it on their side

keiiii3257:
close https://github.com/kiat/codeAssist/issues/314
Instructor are able to to set publish/unpublish state of an assignment.

Expected behavior:
Instructor:

- The table in the dashboard and assignment tab should show all assignment
- Instructor dashboard assignment count all assignments
need

- Screenshot 2026-07-14 at 5 58 20 PM
- Able to set the publish/unpublish state of an assignment
- A new indicator of publishing status for assignments in the dashboard and assignment tab's table

Student:

- Student cannot see unpublished assignments.
- Student dashboard assignment count only visible assignments.

Frontend

- Hide unpublished assignments from student assignment list
- Hide scheduled assignments from students until their release date
- Count only visible assignments on the student dashboard
- Update instructor publish indicator to reflect scheduled release state

Backend

- Enforce assignment visibility on /get_course_assignments.
- Use the authenticated user’s course role (require_course_role) instead of accepting a user_id query parameter.
- Return all assignments for instructors and TAs.
- Return only published and released assignments for students.
- Updated backend unit tests to cover instructor and student behavior with the new authentication flow.

Limitations / Future Enhancement (Separate PR)

- Scheduled assignments are still displayed as Unpublished instead of Scheduled in:
- frontend/src/pages/instructor/dashboard/index.js
- frontend/src/pages/instructor/assignments/Assignments.js
- Scheduled assignments are not automatically published once their release time is reached. Automatic release scheduling is outside the scope of this PR.


DarioDeSilva (changes requested):
Requesting changes, mainly because of finding #1 below.

**1. Assignment visibility filter ignores extensions (backend/routes/course.py:719-732)**

The new student-only filter only checks `Assignment.published` and `Assignment.published_date`. It never looks at `AssignmentExtension.release_date_extension`, which `code_editor.py` and `submission.py` already treat as the authoritative per-student release date. That makes this bug bidirectional:

- Early access: a student with an extension that moves their release date earlier still won't see the assignment in their list, even though a submission would be accepted if they reached it directly.
- Delayed access: a student with an extension that pushes their release date later will still see the assignment as available, but their submission gets rejected as premature.

Can we join/query `AssignmentExtension` here the same way the submission paths do?

**2. Frontend filter has the same blind spot (frontend/src/common/assignmentVisibility.js:16-21, pages/assignments/index.js:99-107)**

`isAssignmentVisible` also only checks `published`/`published_date`, and it's applied before the effect that fetches `/get_extension` and overlays `release_date_extension`. So even after fixing #1 on the backend, this client-side filter would strip the assignment before the extension logic runs. Same fix needed here, or reorder so the filter runs after extension data is merged in.

**3. Test doesn't actually check the filter (backend/test/unit/test_course.py:743-761)**

`test_get_course_assignments_student_only_receives_visible_assignments` mocks `.filter()` and only asserts it was called once, it never checks what was passed in. A regression on the wrong column, a dropped `or_`, or `<=` flipped to `<` would still pass. Worth asserting on the actual filter args, or adding an integration test against a real DB, especially since this test would not have caught #1.

Smaller, non-blocking:

- Duplicate import of `require_course_role` in course.py (already imported at line 23, re-imported at line 30).
- `user_id` is now threaded through several frontend call sites (dashboard, Assignments.js, DuplicateAssignmentModal.js, instructor dashboard) but the backend never reads it, auth is session-based via `require_course_role`. Looks vestigial, and it's applied inconsistently since courseSettings/index.js calls the same service without it. Worth removing unless there's a reason to keep it.
- In `assignmentVisibility.js`, an unparseable `published_date` makes `moment().isBefore(InvalidDate)` evaluate false, so the assignment reads as visible by default. Low risk given the field comes from a Postgres timestamp, but worth a guard.

Overall the server-side role filtering is the right approach for the underlying problem (students seeing unpublished/future assignments), just needs the extension case handled before merge.

Issue #147, opened 2025-02-18
Optimize Time to Handle Assignment Submission


manafasif:
Currently, the autograder takes a long time to handle an assignment submission. This is likely due to how docker containers are managed.

Currently, containers remain stopped, and are only started up when an assignment is submitted.

Features to implement:
1. Switch to lightweight container image

2. Keep assignment container running at all times - or implement an intelligent system to evaluate when to keep it running or not

3. Determine any other ways to minimize time to process each submission.

manafasif:
Implement asynchronous grading. Submitting assignment should return an OK/Fail response, triggering a "grading" job and closing the upload window. 

The job should be handled by a separate thread (from main thread), which updates some in-memory variables relating to the status/progress of the job.

Frontend can continously poll a "status" endpoint to read the status of the job until it finishes. 

This is essential, as current system will fail to handle a long-running autograder, or multiple submissions at once, which will slow the system.

DarioDeSilva:
Related to #148, which was split off from a comment on this issue to track the async job processing piece specifically. Still open — grading is fully synchronous today, no job queue or status endpoint exists.

DarioDeSilva:
Container reuse is in #375, and async grading is tracked in #148. Once #375 merges I think this can close, since the only item left is the lightweight image, which is minor.

PR #375, opened 2026-08-12
Persistent Assignment Containers and Submission Archiving

### LINKED ISSUE #87: Docker Setup on Assignment Creation Cleanup (Prevent New Docker Container Creation on Each Submission)

RaymondZhu3:
Closes #87.
Addresses the persistent-container item in #147.

Re-opens the work from PR #347, which was lost when the branch it targeted was force-pushed. This PR restores it and adds the filesystem work that was in progress at the time.

## Persistent per-assignment containers

Previously every submission created and destroyed its own Docker container. Now each assignment owns one persistent container that is reused across submissions.

- `Assignment.container_id` column added and Alembic migration included.
- `get_or_create_assignment_container` reuses the stored container, and recreates it if it was removed externally or if the autograder image changed (tag comparison is normalized so `image` and `image:latest` match).
- The container is created eagerly when the instructor uploads the autograder, so the first student submission does not pay the startup cost.
- `get_container_lock(assignment_id)` serializes container operations per assignment.
- `exec_run` has no native timeout, so calls are wrapped via a module-level `ThreadPoolExecutor` with `future.result(timeout=...)`. On timeout the container is reset and an `InternalProcessingError` is raised.
- Containers are removed when their assignment is deleted, via an `after_delete` listener on `Assignment`.

## Submission file archiving

Submissions already got their own directory on disk, but nothing ever cleaned it up or connected it back to the database row. Both are fixed here.

- One identifier: the `Submission` primary key is now also the on-disk directory name. Previously a separate `uuid4().hex[:8]` named the folder, so nothing on disk could be correlated to a database row.
- After the row commits, `archive_staged_files` moves the submission and its results JSON into `upload_autograder/archive/{assignment_id}/{submission_id}/` and removes the staging directory. `runs/{assignment_id}/submission/` now only holds submissions still being graded; `archive/` only holds submissions that actually committed.
- Failure paths (autograder non-zero exit, exec timeouts) clean up their staging directory instead of leaving it behind.
- Deleting an assignment removes its `runs/` and `archive/` trees. Both the single delete and the bulk course-level delete are covered.
- The background AI-feedback thread is now handed the archived path rather than the staging path, since the staging directory no longer exists by the time the thread runs.
- `migrations/env.py` had a stray `target_metadata = None` shadowing the real `target_metadata = db.metadata` above it, which broke autogenerate. Removed.
- Covers #87's "maintaining a structured file system... to handle different students, their submissions, and results": submissions and results are stored for each assignment and for each submission under `archive/`, and the container's `/autograder/submission` and `/autograder/results` are cleared between runs.

## Testing

- Two unit tests in `test/unit/test_submission.py` cover archiving on both the no-autograder and successful-run paths, asserting files land in the archive and the staging directory is removed.
- Manually verified by this workflow: create assignment → upload autograder → submit → confirm archive contents and empty staging directory → delete assignment → confirm container and both directory trees removed.
- Stress harness (`test/stress/`) was broken by an auth change and is repaired here for `upload_submission` / `add_user`.

## Follow-ups

- Grading is still fully synchronous. The request blocks until the autograder finishes. #148 (async job queue + status endpoint) is unaffected by this PR and remains future work.
- `rerun_submission_autograder` still uses its own temporary `rerun_{uuid}` directory and is untouched here. It needs cleanup-on-completion rather than archiving, since it pulls code already stored in the database and shares the original submission's ID.
- The stress harness cannot clean up its users: `delete_user` requires an admin account and there is no bootstrap path to create one.


RaymondZhu3:
Fixed four of the issues Copilot flagged:

- **Lock race**: `defaultdict` created and inserted in two steps, so concurrent first access could hand two threads different locks. Now a plain dict with `setdefault(assignment_id, threading.Lock())` so every caller gets the same lock.
- **`NotFound` on cleanup**: a container that's already been removed now exits cleanly instead of logging a warning.
- **`_normalize_tag`**: a colon only counts as a tag when it follows the last `/`.
- **Test**: set comparison instead of exact list equality.

Ready for another review.

DarioDeSilva (changes requested):
Thanks for restoring this work and extending it. The container-reuse and archiving design reads well and the migration chain is clean. I have a few things I'd like to sort out before merge, roughly in priority order. Happy to talk through any of them.

## Would like to address before merge

**1. `course.py` `delete_all_assignments`: filesystem and container teardown runs before the DB delete transaction.**
The per-assignment loop stops/removes containers and `shutil.rmtree`s `runs/{aid}` and `archive/{aid}`, then the bulk `RegradeRequest` / `Submission` / `Assignment` delete runs. If that delete raises (FK error, lock timeout, dropped connection), the `except` calls `db.session.rollback()`. The assignment and submission rows survive, but the archived student submissions and results JSON are already gone and the containers are destroyed. Could we move the destructive filesystem/container work to after a successful commit, or collect the paths/ids first and only act once the transaction succeeds?

**2. `models.py` `cleanup_assignment_container`: blocking Docker I/O inside an `after_delete` mapper event.**
`container.stop()` (10s default grace) and `container.remove(force=True)` run over the Docker socket during flush, while the DB transaction and row locks are still open. SQLAlchemy recommends against external I/O in mapper-level flush events, and a hung Docker daemon would stall the request while holding locks. Could this move to after the commit in the route handler instead of a mapper event?

**3. `submission.py`: staging directory can leak on error paths.**
The staging setup isn't wrapped in try/finally, so a few paths leave `runs/{aid}/submission/{sid}/` behind: the pre-run `rm -rf` exec returning non-zero, `put_archive` raising `docker.errors.APIError`, and `get_or_create_assignment_container` raising. That breaks the stated invariant that `runs/` only holds in-progress submissions. A try/finally around staging would cover every exit path.

## Worth fixing

**4. `exec_run_with_timeout` can't cancel a wedged exec.**
When `future.result(timeout=...)` fires, the submitted `container.exec_run` keeps running on its pool thread. With 8 workers, repeated incidents over the process lifetime could exhaust the pool and make `_exec_pool.submit()` block for all assignments. Could you confirm the container reset reliably unblocks the streaming exec socket? A dedicated executor per assignment, or a hard kill of the exec, might be worth considering.

**5. `reset_assignment_container` only catches `NotFound` from `containers.get()`.**
In the `TimeoutExpired` handler, any other `docker.errors.APIError` (daemon busy, Docker 500) propagates before `db.session.add(failed_submission)` and `raise SubmissionTimeoutError(...)`, so the student gets an opaque 500 and no timeout submission row is recorded. Catching `APIError` here and still recording the failed submission would be more robust.

## Minor, fine to defer with a note

6. `_container_locks` / `get_container_lock` never remove entries, including on assignment deletion, so the dict grows over time. A cleanup in the delete path would close it.

7. Archiving and container reuse were added only to `upload_submission`. `code_editor.py::submit_code` still creates/destroys a container per submission and never archives, and it hands `async_get_ai_feedback` a staging path this PR's model assumes is gone. Either bringing that path in line or noting it as deliberately out of scope in the PR description would help.

8. `archive_staged_files` leaves empty scaffolding dirs (`runs/{aid}/submission/`, per-student `results/` dirs) behind after each submission, which slightly contradicts the "only holds in-progress" claim.

9. `get_or_create_assignment_container` persists `container_id` with a bare `db.session.commit()`. It works today because call sites have clean sessions, but as a shared helper it would flush any unrelated pending changes. A targeted flush of the assignment row would be safer.

RaymondZhu3:
Thanks for the review, addressed all nine requested changes:

1. Teardown before the delete transaction: ids and container ids saved first; containers and the `runs/` and `archive/` trees are destroyed only after the commit succeeds. A failed delete now rolls back with the archived submissions still there.
2. `after_delete` mapper event: both listeners removed (the directories one had the same bug in the single-delete path) and called from the routes after commit instead. No coverage lost, since `delete_course` rejects courses that still have assignments, so those two routes are the only paths that destroy an Assignment.
3. Staging leak: try/finally around everything after `stage_submission`, covering all three paths you listed; the four per-branch rmtree calls are gone. Most of that file's diff is the re-indentation, so `git diff --ignore-all-space` is the readable view.
4. Wedged exec: replaced the pool with a `docker exec` subprocess, which is actually killed on timeout, so there's no pool left to exhaust. Helper execs raise a distinct `ContainerExecTimeout`, since sharing `subprocess.TimeoutExpired` with the autograder run would have let a container-cleanup timeout get recorded as a student's failed submission.
5. `reset_assignment_container`: now catches `DockerException` too.
6. `_container_locks`: `discard_container_lock` called from both delete paths, with a guard so the setdefault/pop pair is safe.
7. `code_editor.py`: out of scope, will create a follow-up issue on this. On the staging path it passes to `async_get_ai_feedback`: it stages to its own directory and never cleans up, so that path still exists and nothing here breaks it.
8. Empty scaffolding: removed with `os.rmdir`, which refuses non-empty directories, so a concurrent submission still staging keeps its parent alive.
9. Bare `db.session.commit()`: `container_id` now written on its own connection.

Ready for another review.

RaymondZhu3:
Thanks for the review, addressed all nine requested changes:

1. Teardown before the delete transaction: ids and container ids saved first; containers and the `runs/` and `archive/` trees are destroyed only after the commit succeeds. A failed delete now rolls back with the archived submissions still there.
2. `after_delete` mapper event: both listeners removed (the directories one had the same bug in the single-delete path) and called from the routes after commit instead. Nothing else deletes an Assignment: `delete_course` rejects courses that still have assignments, so those two routes are the only paths.
3. Staging leak: try/finally around everything after `stage_submission`, covering all three paths you listed; the four per-branch rmtree calls are gone. Most of that file's diff is just the re-indentation.
4. Wedged exec: replaced the pool with a `docker exec` subprocess, which is actually killed on timeout, so there's no pool left to exhaust. Helper execs raise a distinct `ContainerExecTimeout`, since sharing `subprocess.TimeoutExpired` with the autograder run would have let a container-cleanup timeout get recorded as a student's failed submission.
5. `reset_assignment_container`: now catches `DockerException` too.
6. `_container_locks`: `discard_container_lock` called from both delete paths, with a guard so the setdefault/pop pair is safe.
7. `code_editor.py`: out of scope, will create a follow-up issue on this. The staging path it hands `async_get_ai_feedback` still exists, since `code_editor` stages to its own directory and never cleans it up, so nothing here breaks it.
8. Empty scaffolding: removed with `os.rmdir`, which refuses non-empty directories, so a concurrent submission still staging keeps its parent alive.
9. Bare `db.session.commit()`: `container_id` now written on its own connection.

Ready for another review.

DarioDeSilva:
## Review: Persistent Assignment Containers and Submission Archiving

Thanks for restoring this work and pushing it forward. The archiving side in particular is a real improvement, and the "tear down only after a successful commit" discipline is done carefully and well tested. Most of the comments below are about the persistent container half, which changes the isolation model of the autograder in a few ways that I think need another look before this lands.

For context on what was checked: the migration chain is clean (single head `a3f92c1e7b04`, linear onto `e6f7a8b9c0d2`), the `env.py` `target_metadata` fix is correct, there is no circular import from the new `routes.submission` imports in `assignment.py` and `course.py`, and the three touched test files all pass (216 tests).

### Blocking / high severity

**1. The persistent container is not reset between students beyond the submission and results folders**

`submission.py` around line 530 scrubs the container between runs with:

```python
exec_run_with_timeout(container, "rm -rf /autograder/submission/* /autograder/results/*")
```

Everything else a student's code touched carries over to the next student:

- `rm -rf .../*` does not match dotfiles, so things like `.pth` files, a dropped `sitecustomize.py` or `conftest.py`, or a `.pytest_cache` survive into the next run.
- `/autograder/source/` (the grader harness itself), pip packages installed at grade time, `/tmp`, `/root`, environment variables, and `__pycache__` are never reset. Since many autograders import or exec the submission, a student who writes to `/autograder/source/run_autograder` or the test files can influence the score of every student graded after them on that container, or read the next student's file while it briefly sits in `/autograder/submission/`.
- On the path where the autograder exits non-zero (not a timeout), the container is left running as-is. Any background process or thread the student's code spawned keeps running, and `rm -rf` does not kill processes. The old per-submission container was destroyed on every exit path, so none of this was reachable.

Suggested direction: reset the whole `/autograder/submission` and `/autograder/results` directories including dotfiles (remove and recreate the dirs, or `find <dir> -mindepth 1 -delete`), restore or read-only-mount `/autograder/source` so a submission cannot modify the harness, and run the student code as a non-root user without write access to `/autograder/source`. It is also worth a short note in `docs/ai/` that the trust boundary between students changed with this design.

**2. `_container_locks` only serializes within one process**

`submission.py` line 31. The lock is a module level dict, so it works under the current `flask run` single process setup but not under gunicorn or uWSGI with more than one worker. With multiple workers, two processes can enter `get_or_create_assignment_container` for the same assignment at the same time and grade concurrently in the one shared container: one run's `rm -rf /autograder/submission/*` and `put_archive` land in the middle of the other run, so a student gets a wrong score or a spurious "Failed to grade submission". The old design gave each submission its own uniquely named container, so this coupling did not exist.

`code_editor.py` around line 44 already documents this same limitation for its in-process rate limiter, but here the failure mode is a corrupted grade rather than an imprecise counter. This needs either a cross-process lock (a Postgres advisory lock keyed on the assignment id, or a file lock) or a clearly documented "single process only" constraint on how the backend is deployed.

**3. The lock is held for the entire grading run, which serializes all submissions to an assignment**

`submission.py` line 523. `get_container_lock(assignment.id)` wraps the autograder `subprocess.run(..., timeout=assignment.autograder_timeout)` (default 300s) plus the results read, and the HTTP request blocks the whole time. So all submissions to one assignment run strictly one at a time, each holding a WSGI worker thread for up to `autograder_timeout`. At a deadline with, say, 40 students and a 60s autograder, the 40th request waits around 39 minutes and most clients or proxies will have timed out long before that. This is a throughput regression from the previous per-submission containers, which ran in parallel.

Async grading (#148) being out of scope here is a reasonable call. The point is more that persistent containers force this serialization, so #148 becomes more important as a follow-up, and it would help to say that in the PR description or a code comment so the tradeoff is visible.

**4. Re-uploading a fixed autograder may keep grading against the old image**

`submission.py` around line 240. `image_name` is a constant tag per assignment (`autograder-<id>`), and the staleness check compares image tags rather than image IDs. It currently works only because rebuilding the image orphans the old tag, so `container.image.tags` comes back empty and the container is recreated. That breaks if the rebuild is a Docker cache hit (identical build context produces the same image ID and the tag never moves), in which case students keep hitting the stale grader after the instructor thought they fixed it.

Suggested fix: have `upload_assignment_autograder` call `reset_assignment_container(assignment)` before the eager recreate, and/or compare `container.image.id` against the freshly built image's ID rather than tags. A test for "rebuild autograder, next submission uses the new image" would lock this down.

### Medium severity

**5. The 409 conflict recovery path never starts a stopped container**

`submission.py` lines 272 to 288. When `containers.run(name=...)` raises a 409, the code fetches the existing container by name and returns it. The `elif container.status != "running": container.start()` branch hangs off `if container is None`, so it is not reachable from the 409 path. Combined with finding 6, this can wedge an assignment:

- `reset_assignment_container` logs a Docker error and sets `container_id = None` without removing the container ("clearing container_id anyway"), or a daemon restart leaves `assignment_container_<id>` in the `exited` state.
- Next submission: `container_id` is NULL, `containers.run(name=...)` raises 409, the code fetches the stopped container, persists its ID, and immediately runs `docker exec` against it. That exits non-zero and the student gets a 500 ("Failed to cleanup container").
- It only recovers on a later request that happens to take the top branch and calls `container.start()`.

Fix: in the 409 branch, `reload()` and start the container if it is not running, sharing the logic with the `elif`.

**6. A container created without a persisted ID is leaked on assignment delete**

`api/models.py` lines 52 to 53. `cleanup_assignment_container` returns early when `container_id` is falsy. After the scenario in finding 5 (ID cleared but container not removed), deleting the assignment calls `cleanup_assignment_container(None, ...)`, which is a no-op, and `assignment_container_<id>` keeps running with nothing referencing it. `cleanup_assignment_container` should also try `containers.get(f"assignment_container_{assignment_id}")` by name and remove that if present.

**7. Teardown races with an in-flight submission**

`assignment.py` around line 290 and the bulk delete in `course.py`. Both call the cleanup helpers and pop the lock without holding it. If a submission is grading concurrently:

- its `docker exec` gets killed when the container is removed, so the student sees an opaque 500, or
- it finishes and reaches `db.session.add(new_submission); db.session.commit()` at `submission.py` line 654 against an assignment that no longer exists. That final commit has no try/except, unlike every other commit in the function, so it raises an uncaught `IntegrityError` and leaves the session in a bad state, or
- `get_or_create_assignment_container` recreates a container keyed to an assignment that was just deleted, and finding 6 then leaks it.

Deleting an assignment while students are actively submitting is rare, but taking `get_container_lock(assignment_id)` around the teardown and wrapping that success-path commit would close it.

**8. `reset_assignment_container` commits in a `finally`, which can swallow the failed-submission record**

`submission.py` lines 345 to 347. On `subprocess.TimeoutExpired`, `upload_submission` calls `reset_assignment_container(assignment)` before it builds `timeout_result` and writes the failed `Submission` row. If the `db.session.commit()` in that `finally` raises (DB briefly unavailable, or the session already in a bad state), the exception propagates out of the except handler, the student gets a plain 500, and no timeout submission is recorded. That is the outcome the swallow-and-log design in the docstring is meant to prevent. Catching broadly inside `reset_assignment_container`, including around the commit, would make it match its stated contract.

**9. `json.loads(results_json_content)['score']` is unguarded**

`submission.py` lines 645 to 646. A malformed autograder result, or one missing the `score` key, raises `KeyError` or `JSONDecodeError` here. That becomes a raw 500 with no `Submission` row, and the `finally` deletes the staged code, so nothing is kept. The `rerun_submission_autograder` path already guards this with a `try/except json.JSONDecodeError` around line 1216, so it would be good to make the two consistent.

### Design and operational notes

**10. The on-disk archive duplicates data that is already in the database, with no retention limit**

Every submission's code and `results.json` are now stored permanently under `archive/{assignment_id}/{submission_id}/`, in addition to the identical bytes in `Submission.student_code_file` and `Submission.results`. Cleanup only happens on assignment deletion. For a busy course this is a continuously growing second copy of everything. If the database is the system of record, it would be worth deciding whether the archive needs to be permanent or could have a TTL.

**11. Idle containers accumulate with no reaping**

Each assignment that has an autograder keeps a container running `tail -f /dev/null` indefinitely, created eagerly on autograder upload. A semester of assignments means dozens to hundreds of idle containers holding memory and PIDs, with no eviction. The old model only ran containers during grading. Some kind of idle reap (remove after N minutes unused, recreate on demand) would bound this.

**12. The two submission paths have diverged**

`code_editor.py::submit_code` still uses ephemeral per-submission containers, the old `uuid4().hex[:8]` staging directory names, and does not archive, so editor submissions still leak their staging directories, which is the bug this PR fixes for uploads. The "submission primary key is the on-disk directory name" property only holds for the upload path. Either bring `submit_code` onto the shared helpers or document the split in `docs/ai/`.

**13. `cleanup_assignment_container` and `cleanup_assignment_directories` in `api/models.py`**

These pull `docker`, `shutil`, and `logging` into the models module and re-create the Docker client with `docker.from_env()` instead of the existing `get_docker_client()` singleton. The docstrings are helpful, but this reads more like it belongs next to the other container helpers in `routes/submission.py` or a small service module.

### Minor

- `submission.py` has several `os.chdir(current_dir)` calls (lines 532, 557, 563, 609, 619, 622, 659) but nothing in the function changes the working directory, so they are no-ops. `os.chdir` is also process global, which is risky in a threaded handler. Worth removing.
- `submission.py` lines 627 to 644 write the results JSON to a host file, then read it back with `open(...).read()` for the DB column, then `json.loads` it two or three times. The in-memory `results_json_content` string already has everything.
- The unique `results_<submission_id>.json` name and the `mv` in the exec command guard a race that the in-process lock already prevents, and that the cross-process case (finding 2) is not protected from anyway. It adds a bit of complexity for little gain.
- No-autograder path: `runs/<assignment_id>/<student_id>/results/` is created at line 484 but never used and never pruned, so these empty directories accumulate one per student per assignment until the assignment is deleted.

### Tests

The archiving invariants are well covered (empty-scaffolding pruning, keeping a parent when another submission is still staging, commit ordering for both delete routes including the rollback case, `ContainerExecTimeout` vs `subprocess.TimeoutExpired`). A few gaps:

- No test that rebuilding the autograder image recreates the container (finding 4), which is the riskiest untested path.
- `get_or_create_assignment_container` is mocked out in every `upload_submission` test, so the tag comparison, the separate-connection `container_id` write, and the `db.session.expire` are never exercised.
- `test_upload_submission_success_archives_submission_and_results` does not stub `threading.Thread`, so it spawns the real `async_get_ai_feedback` against the test database. It passes now but it is a likely source of future flakiness. Worth patching the thread.

### What is good

- The "clean up Docker and the filesystem only after a successful commit, never before and never in a flush hook" rule is the right call, clearly explained in the docstrings, and tested from both delete routes.
- `exec_run_with_timeout` using a subprocess that actually kills the exec, rather than `exec_run` on a pool thread, is the right mechanism, and keeping `ContainerExecTimeout` distinct from `subprocess.TimeoutExpired` so a stuck helper is not recorded as a student timeout is a nice detail.
- Using `Submission.id` as the on-disk directory name is a clear improvement over the disconnected `uuid4().hex[:8]`.
- The `env.py` `target_metadata` fix is correct and unblocks autogenerate.
- Staging cleanup in the `finally` genuinely closes the leak on the upload path.

RaymondZhu3:
Thanks for the review, addressed findings 1-9 plus the minor items and test gaps:

1. Container reset: `rm -rf dir/*` skips dotfiles, nothing restored `/autograder/source`, and it can't kill processes. `reset_container_workspace` now reaps with `kill -9 -1`, restores the harness from a baseline snapshot taken at container creation, and clears both directories with `find -mindepth 1 -delete`. A failed reset raises instead of grading against leftovers, and the non-zero-exit path discards the container like the timeout path.
2. Cross-process lock: took the advisory lock rather than documenting a single-process constraint, since our deploy plan uses gunicorn. `assignment_container_lock` layers `pg_advisory_lock` over the in-process lock on its own connection, since a session-level lock drops when its connection commits. Key is blake2b, not `hash()`, which `PYTHONHASHSEED` randomizes per process. Falls back to the in-process lock on sqlite.
3. Serialization: agreed, inherent to one container per assignment. Noted in the `assignment_container_lock` docstring with a pointer to #148.
4. Stale image: `_container_image_is_stale` compares image IDs instead of tags, and `upload_assignment_autograder` resets the container before recreating.
5. 409 path: start logic moved out of the `if container is None` chain so it runs for either branch.
6. Leaked container: `cleanup_assignment_container` tries the recorded id, then `assignment_container_<id>`, de-duplicating on `container.id`.
7. Teardown race: both delete routes hold `assignment_container_lock` across teardown, with the lock popped after release. Success-path commit wrapped like the others.
8. `reset_assignment_container`: catches `Exception`, and the commit is inside a try with a rollback, so a commit failure can't lose the timeout row.
9. Results parse: parsed once and guarded. Unusable output records a score-zero `completed=False` row and archives the staged code before raising, since it's the student's only record of the attempt. Shared with the timeout path via `_record_failed_submission`.

Minor items: removed the seven no-op `os.chdir` calls; results column built from the string already in memory; in-container `mv` gone, with the host-side file keeping the unique name; per-student `results/` created only on the autograded path. Also factored out `discard_container` and `commit_or_rollback` for the two repeated error-handling shapes, leaving the occurrences in `rerun_submission_autograder`, `test_autograder_submission` and `code_editor.py` alone to keep the diff to the reviewed findings.

10-13 taking as follow-ups, will file issues. 10 needs a call on whether the DB or the disk is the system of record before deciding if the archive should expire; 11 needs an eviction policy plus recreate-on-demand; 12 means touching the second submission path; 13 is a pure refactor, and `models.py` can't call `get_docker_client()` as-is without a circular import.

Tests: 502 pass after merging `main` (473 on the branch, up from 458). Added coverage for the rebuild recreating the container including the cache-hit case, `get_or_create_assignment_container` unmocked, the reset, the advisory lock, the `cleanup_assignment_container` name fallback, and malformed results.

Merged `main`. Ready for another review.

DarioDeSilva:
Followed up on the latest commit (`b9fe007740`, "Address requested changes: container reset isolation, cross-process locking, container lifecycle"). Most of the round two items look resolved: workspace reset now handles dotfiles and restores `/autograder/source` from a snapshot, the advisory lock covers cross-process serialization, image staleness compares IDs instead of tags, the 409 recovery path starts a stopped container, container cleanup on delete also matches by name, teardown for delete_assignment/delete_all_assignments now holds the lock, and malformed results.json is now guarded.

Two things came up that are new or still open and worth addressing before merge.

**1. `reset_container_workspace` failures escape the exception handling in `upload_submission` (submission.py, around line 724)**

`reset_container_workspace` raises `InternalProcessingError` directly when the reset script exits non-zero, but the surrounding try only catches `ContainerExecTimeout` and `subprocess.TimeoutExpired`. If the reset fails for any other reason (leftover root-owned file, permission error on the find/delete), that exception propagates without ever calling `reset_assignment_container`, so `assignment.container_id` keeps pointing at the broken container. Every following submission to that assignment hits the same reset failure and 500s until an instructor happens to re-upload the autograder. Suggested fix: catch `InternalProcessingError` from this call too (or have `reset_container_workspace` call `reset_assignment_container` itself before raising).

**2. Autograder non-zero exit deletes the submission with no record and no archive (submission.py, around line 781)**

The `exec_proc.returncode != 0` branch now calls `reset_assignment_container` and raises, but unlike the timeout and malformed-results branches it never calls `_record_failed_submission` or `archive_staged_files`. The new unconditional `finally: shutil.rmtree(submissions_dir, ignore_errors=True)` then deletes the staged file. Net effect: if a student's run_autograder script itself exits non-zero, the submission vanishes completely, no Submission row and no archived copy. This is a regression from the old code, which at least left the staged file on disk. Should probably route through `_record_failed_submission` the same way the other two failure paths do.

**One more from the earlier review that's still open**

`duplicate_assignment` (assignment.py, around line 236) still copies `container_id` and `autograder_image_name` straight through via `AssignmentSchema().dump(old_assignment)`, only overriding id/name/course_id. Duplicating an assignment that already has a live persistent container hands the new assignment the same container_id, and since autograder_image_name is copied too the staleness check won't catch it, so the first submission to the new assignment reuses the old assignment's container directly. create_assignment already nulls container_id for this reason, duplicate_assignment should do the same.

